PR #1, opened 2016-05-24
[JENKINS-35110] Add exws definitions and templates as global Jenkins …

alexsomai:
…configs. Add exws node property.

CC @oleg-nenashev @martinda


pom.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>plugin</artifactId>
+        <version>2.5</version>
```
oleg-nenashev: better to pick the latest version

pom.xml
```diff
@@ -0,0 +1,74 @@
...
+        <relativePath/>
+    </parent>
+
+    <groupId>org.jenkins-ci.plugins.workflow</groupId>
```
oleg-nenashev: I would vote against "workflow" in the group ID, because later the plugin can be used for other item types

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -0,0 +1,94 @@
...
+import hudson.model.Descriptor;
+import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.StaplerRequest;
+
+import javax.annotation.Nonnull;
+
+/**
+ * @author Alexandru Somai
```
oleg-nenashev: If you create Javadoc, It's better to put a meaningful class description

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -0,0 +1,94 @@
...
+        public boolean configure(StaplerRequest req, JSONObject formData) throws FormException {
+            req.bindJSON(this, formData);
+            save();
+            return super.configure(req, formData);
+        }
+
+        @Nonnull
+        @Override
+        public String getDisplayName() {
+            return "Disk";
```
oleg-nenashev: Good practice: use Jenkins localization features: https://wiki.jenkins-ci.org/display/JENKINS/Internationalization

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -0,0 +1,94 @@
...
+
+        public void setDiskId(String diskId) {
+            this.diskId = diskId;
+        }
+
+        public String getName() {
+            return name;
+        }
+
+        public void setName(String name) {
```
oleg-nenashev: It's a bad idea to provide such setter API, because there is no automatic save of the instance after value modification.
My recommendation would be to remove setters from the public space if there is no real need. 🐜

src/main/java/org/jenkinsci/plugins/ewm/definitions/Template.java
```diff
@@ -0,0 +1,97 @@
...
+
+        public String getDiskRefId() {
+            return diskRefId;
+        }
+
+        public void setDiskRefId(String diskRefId) {
+            this.diskRefId = diskRefId;
+        }
+
+        public String getLabel() {
```
oleg-nenashev: For such methods it's definitely makes sense to define the behavior for empty values. I would nullify them and add a CheckForNull annotation to the method

oleg-nenashev: For such methods it's definitely makes sense to define the behavior for empty values. I would nullify them and add a CheckForNull annotation to the method

src/main/resources/org/jenkinsci/plugins/ewm/nodes/DiskNode/help-localRootPath.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    The mounting point to the physical disk.
+</div>
```
oleg-nenashev: Good practice: always add line break to the file end

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -0,0 +1,78 @@
...
+        public DescriptorImpl() {
+            super(ExwsAllocateExecution.class);
+            load();
+        }
+
+        @Override
+        public boolean configure(StaplerRequest req, JSONObject formData) throws FormException {
+            req.bindJSON(this, formData);
+            save();
+            return super.configure(req, formData);
```
oleg-nenashev: It likely leads to the double save operation

oleg-nenashev:
Looks good for the prototype stage.

oleg-nenashev:
Just for the future, it's better to avoid commits including multiple changes. It's difficult to review diffs in such case

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
alexsomai: I've changed this on #3

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -0,0 +1,77 @@
...
+
+        public FormValidation doCheckName(@QueryParameter String value) {
+            return validateRequired(value);
+        }
+
+        public FormValidation doCheckPhysicalPathOnDisk(@QueryParameter String value) {
+            return validateRequired(value);
+        }
+
+        @Nonnull
```
alexsomai: If I remove this, my IDE gives me a warning saying:
_Not annotated method overrides method annotated with @\Nonnull_

src/main/resources/org/jenkinsci/plugins/ewm/nodes/DiskNode/help-localRootPath.html
martinda: The user may not have a physical disk (it might be a symbolic link again). So perhaps the help should say "The mounting point from this node to the disk"

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
oleg-nenashev: Which IDE do you use?

oleg-nenashev:
👍 for merging

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
alexsomai: IntelliJ. Anyway, I can disable that warning if you are saying that the annotation is not needed.

PR #3, opened 2016-05-30
[JENKINS-35192] Implement the exwsAllocate step for the upstream job

alexsomai:
This is the implementation of the exwsAllocate step for the upstream job.

CC @oleg-nenashev  @martinda

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -1,33 +1,65 @@
 package org.jenkinsci.plugins.ewm.steps;
 
 import com.google.inject.Inject;
+import hudson.FilePath;
 import hudson.model.Run;
-import org.jenkinsci.plugins.ewm.definitions.DiskPool;
+import hudson.model.TaskListener;
+import org.jenkinsci.plugins.ewm.definitions.Disk;
+import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
+import org.jenkinsci.plugins.ewm.strategies.DiskAllocationStrategy;
+import org.jenkinsci.plugins.ewm.strategies.MostUsableSpaceStrategy;
 import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
 import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
 
-import java.util.List;
+import java.io.File;
 
 /**
- * TODO - To be added when I'll implement the step
+ * The execution of the {@link ExwsAllocateStep}.
  *
  * @author Alexandru Somai
  */
-public class ExwsAllocateExecution extends AbstractSynchronousNonBlockingStepExecution<Void> {
+public class ExwsAllocateExecution extends AbstractSynchronousNonBlockingStepExecution<ExternalWorkspace> {
 
     @Inject(optional = true)
-    private transient ExwsAllocateStep exwsAllocateStep;
+    private transient ExwsAllocateStep step;
 
     @StepContextParameter
     private transient Run<?, ?> run;
+    @StepContextParameter
+    private transient TaskListener listener;
 
     @Override
-    protected Void run() throws Exception {
-        List<DiskPool> diskPools = exwsAllocateStep.getDescriptor().getDiskPools();
-        // TODO use globally defined disk pools
-        System.out.println(diskPools);
+    protected ExternalWorkspace run() throws Exception {
+        if (step.getUpstream() == null) {
+            // this is the upstream job
 
-        // TODO implement method
-        return null;
+            DiskAllocationStrategy allocationStrategy = new MostUsableSpaceStrategy(step.getDiskPoolId(), step.getDescriptor().getDiskPools());
+            Disk disk = allocationStrategy.allocateDisk();
+
+            String pathOnDisk = computePathOnDisk(disk.getPhysicalPathOnDisk());
+            ExternalWorkspace externalWorkspace = new ExternalWorkspace(disk.getDiskId(), pathOnDisk);
+
+            listener.getLogger().println(String.format("Selected disk id is: %s", externalWorkspace.getDiskId()));
+            listener.getLogger().println(String.format("The path on disk is: %s", externalWorkspace.getPathOnDisk()));
+
+            return externalWorkspace;
+        } else {
+            // this is the downstream job
+            // TODO implement
+            return null;
+        }
+    }
+
+    /**
+     * Computes the path to be used on the physical disk.
+     * The computed path is like: physicalPathOnDisk/$JOB_NAME/$BUILD_NUMBER. Where $JOB_NAME also includes all the
+     * folders, if Folders plugin is in use.
+     *
+     * @param physicalPathOnDisk the physical path on the disk
+     * @return the computed file path on the physical disk
+     */
+    private String computePathOnDisk(String physicalPathOnDisk) {
+        FilePath diskFilePath = new FilePath(new File(physicalPathOnDisk));
+        return new FilePath(diskFilePath, run.getParent().getFullName() + "/" + run.getNumber()).getRemote();
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -25,6 +25,8 @@ import static java.lang.String.format;
  */
 public class ExwsAllocateExecution extends AbstractSynchronousNonBlockingStepExecution<ExternalWorkspace> {
 
+    private static final long serialVersionUID = 1L;
+
     @Inject(optional = true)
     private transient ExwsAllocateStep step;
 
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -5,6 +5,7 @@ import hudson.FilePath;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import org.jenkinsci.plugins.ewm.definitions.Disk;
+import org.jenkinsci.plugins.ewm.definitions.DiskPool;
 import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
 import org.jenkinsci.plugins.ewm.strategies.DiskAllocationStrategy;
 import org.jenkinsci.plugins.ewm.strategies.MostUsableSpaceStrategy;
@@ -12,6 +13,7 @@ import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepEx
 import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
 
 import java.io.File;
+import java.util.List;
 
 /**
  * The execution of the {@link ExwsAllocateStep}.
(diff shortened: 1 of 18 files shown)


src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -33,11 +35,26 @@ protected ExternalWorkspace run() throws Exception {
...
+
+            List<DiskPool> diskPools = step.getDescriptor().getDiskPools();
+            DiskAllocationStrategy allocationStrategy = new MostUsableSpaceStrategy(diskPoolId, diskPools);
             Disk disk = allocationStrategy.allocateDisk();
 
-            String pathOnDisk = computePathOnDisk(disk.getPhysicalPathOnDisk());
-            ExternalWorkspace externalWorkspace = new ExternalWorkspace(disk.getDiskId(), pathOnDisk);
+            String physicalPathOnDisk = disk.getPhysicalPathOnDisk();
+            if (physicalPathOnDisk == null) {
+                throw new Exception("Physical path on disk was not provided in the Jenkins global config");
```
martinda: This is optional, but you could enhance the error message by telling the user the `diskPoolId` to which this error applies.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -33,11 +35,26 @@ protected ExternalWorkspace run() throws Exception {
...
+            }
+            String diskId = disk.getDiskId();
+            if (diskId == null) {
+                throw new Exception("Disk ID was not provided in the Jenkins global config");
```
martinda: Same here, you could enhance the error message by telling the user the `diskPoolId` to which this error applies.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: Yes indeed, this can be optional.
So maybe, if this parameter is not provided, I could fallback to a default path, e.g.: "workspace". And maybe display an info message. WDYT?

martinda: With multiple disk pools, showing the `diskPoolId` in the error message will help with debug.

If we think about a fallback, then I think it would be the node's Remote root directory. If this is too complicated, then I would prefer no fallback, just an error.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -1,33 +1,86 @@
...
+    @StepContextParameter
+    private transient TaskListener listener;
 
     @Override
-    protected Void run() throws Exception {
-        List<DiskPool> diskPools = exwsAllocateStep.getDescriptor().getDiskPools();
-        // TODO use globally defined disk pools
-        System.out.println(diskPools);
+    protected ExternalWorkspace run() throws Exception {
+        if (step.getUpstream() == null) {
```
oleg-nenashev: Maybe this condition os too strict. I can imagine a case when somebody starts an unrelated downstream job, which has its own workspace management logic

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -1,33 +1,86 @@
...
+     * Computes the path to be used on the physical disk.
+     * The computed path is like: physicalPathOnDisk/$JOB_NAME/$BUILD_NUMBER. Where $JOB_NAME also includes all the
+     * folders, if Folders plugin is in use.
+     *
+     * @param physicalPathOnDisk the physical path on the disk
+     * @return the computed file path on the physical disk
+     */
+    private String computePathOnDisk(String physicalPathOnDisk) {
+        FilePath diskFilePath = new FilePath(new File(physicalPathOnDisk));
+        return new FilePath(diskFilePath, run.getParent().getFullName() + "/" + run.getNumber()).getRemote();
```
oleg-nenashev: Full name may contain special symbols. I'm not sure FilePath normalizes all the stuff correctly

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -34,6 +38,15 @@ public String getDiskPoolId() {
         return diskPoolId;
     }
 
+    public String getUpstream() {
+        return upstream;
+    }
+
+    @DataBoundSetter
+    public void setUpstream(String upstream) {
+        this.upstream = upstream;
```
oleg-nenashev: such call requires string normalization (e.g. Blank string => null)

src/main/java/org/jenkinsci/plugins/ewm/strategies/DiskAllocationStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+     * @see DiskAllocationStrategy#allocateDisk(List) ()
+     */
+    @Nonnull
+    public Disk allocateDisk() {
+        DiskPool diskPool = findDiskPool();
+
+        List<Disk> disks = diskPool.getDisks();
+        if (disks.isEmpty()) {
+            String message = String.format("No Disks were defined in the global config for Disk Pool '%s'", diskPoolId);
+            throw new RuntimeException(message);
```
oleg-nenashev: Please do not propagate errors via Runtime exceptions. It always leads to the mishandling of errors. Better to define a new exception type, which would inherit IOException or Jenkins' AbortException

src/main/java/org/jenkinsci/plugins/ewm/strategies/DiskAllocationStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+        DiskPool diskPool = null;
+        for (DiskPool dp : diskPools) {
+            if (diskPoolId.equals(dp.getDiskPoolId())) {
+                diskPool = dp;
+                break;
+            }
+        }
+        if (diskPool == null) {
+            String message = String.format("No Disk Pool ID matching '%s' was found in the global config", diskPoolId);
+            throw new RuntimeException(message);
```
oleg-nenashev: same

src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/config.jelly
```diff
@@ -3,4 +3,7 @@
     <f:entry title="${%Disk Pool Ref ID}" field="diskPoolId">
         <f:textbox/>
     </f:entry>
+    <f:entry title="Upstream" field="upstream">
+        <f:textbox />
```
oleg-nenashev: Missing help file?

src/main/java/org/jenkinsci/plugins/ewm/strategies/DiskAllocationStrategy.java
martinda: @alexsomai, it is the `AbortException` I was thinking about on Monday at the meeting.

alexsomai: @martinda, yeap, I remember that you've said about using Jenkins' specific exception.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: Should I use java.io.File for computing the path?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: This shouldn't have anything to do with the real upstream/downstream definitions in Jenkins. If this is what you are saying.
I'll need this for the [basic-usage](https://github.com/martinda/external-workspace-manager#basic-usage) feature.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
oleg-nenashev: It's even worse. Just keep in mind you may need symbol escaping in the final version

oleg-nenashev:
Just a topic for the design: which module will be responsible for workspace cleanup? By default you create a new workspace for each build, so it's about bloating the storage disks

martinda:
Yes, disk space clean up is something to think about. In my case, I manually manage workspaces with a separate independent Jenkins job that erases all workspaces older than 60 days, while keeping the 10 most recent workspaces forever. This kind of concept, applied locally, could be incorporated to the plugin design.

src/main/java/org/jenkinsci/plugins/ewm/strategies/MostUsableSpaceStrategy.java
```diff
@@ -38,4 +39,21 @@ public Disk allocateDisk(List<Disk> disks) {
...
     }
+
+    /**
+     * Calculates the usable space for the given {@link Disk} entry.
+     *
+     * @param disk the disk entry
+     * @return the usable space for the disk
+     * @throws AbortException if mounting point from Master to Disk is {@code null}
+     */
+    private long retrieveUsableSpace(Disk disk) throws AbortException {
```
oleg-nenashev: I am not a big fan of AbortException, because it is not extensible in the current version

src/main/java/org/jenkinsci/plugins/ewm/strategies/DiskAllocationStrategy.java
```diff
@@ -15,7 +15,7 @@
  */
 public abstract class DiskAllocationStrategy {
 
-    private final String diskPoolId;
+    protected final String diskPoolId;
```
oleg-nenashev: Better to use the getter method

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
oleg-nenashev: Agreed. getUpstream() exists only if you specify the parameter explicitly

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: I've added this issue to be implemented later https://issues.jenkins-ci.org/browse/JENKINS-35302

src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/config.jelly
alexsomai: Added help-upstream.html file.

alexsomai:
Regarding the workspace clean up, I've added a sub-chapter in the design document about it: https://docs.google.com/document/d/1yiisnsR7qg3XEEvch8vocWbitSUCZcoQ-pfzEVFg1eA/edit#heading=h.j6djxdq7n47g
Currently the chapter contains only a sketch of how it could be done, but additional thinking and a good design has to be done on this topic.

alexsomai:
This branch contains all the FindBugs warnings fixed: https://jenkins.ci.cloudbees.com/job/plugins/job/external-workspace-manager-plugin/8/
So, all the other PRs that follow must not have any FindBugs warnings.

alexsomai:
WDYT @oleg-nenashev @martinda, is this ready to be merged into master?

oleg-nenashev:
👍 from me. Show work well for the prototype. This code does not go to the production immediately, so there will be another review iteration later in any case.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -39,6 +39,9 @@ public boolean start() throws Exception {
         System.out.println(templates);
 
         Node node = computer.getNode();
+        if (node == null) {
+            throw new Exception("Computer does not correspond to a live node");
```
martinda: What can the user do to remedy this situation if it occurs? Or what causes this error? I am just thinking, how can the plugin help the user out of this situation if it occurs? I see some possible answers here: http://javadoc.jenkins-ci.org/hudson/model/Computer.html

alexsomai: On the getNode() method, it says 

> Returns:
> null if the configuration has changed and the node is removed, yet the corresponding Computer is not yet gone.

I don't know exactly what the user can do. I see this as a corner case scenario.
This null check and exception I've used them exactly how they are used in the ws step implementation: https://github.com/jenkinsci/pipeline-plugin/blob/1.0/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java#L38

martinda: I was thinking the plugin could tell the user that the node could not be live due to some unexpected conditions: the node might have been take off line, or might have been removed; but maybe I misunderstand what `getNode()` can provide in terms of information.

Another corner condition that I see is with the `Computer` object. If there are zero executors, then it won't exist (null?). Does it make sense for the plugin to cover the case of a non-existent `Computer` instance (even if the user is to blame)?

alexsomai: It's a bit unclear to me what exactly means when `node == null` and how I can reproduce that behavior. I remove a node, and then quickly run the build maybe? 
But yes, the error message may be changed. Is it ok if I change this error message on #6? because on that branch I did the actual implementation for the 'exws' step.

Regarding the `Computer` object, how can I configure a Node with 0 executors? because if I try saving a Node config with 0 executors, it fails with a RuntimeException. IMO, I don't think is the plugin's responsibility to check whether the `Computer` object is null or not.

martinda: Yes, this might be very tricky to test manually, perhaps only a corner case unit test can manage this? Yes you can change the message on the other PR.

Ok, maybe it is not practical to check for a null `Computer` object inside the plugin.

alexsomai:
This PR was created against `upstream/global-config` branch, and not against `upstream/master`.
Closing it in favor of #9

PR #6, opened 2016-06-02
[JENKINS-35244] Implement the exws step

alexsomai:
I'll do smaller commits from now on :)

CC @oleg-nenashev @martinda

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
@@ -12,6 +12,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
 import static hudson.Util.fixEmptyAndTrim;
+import static hudson.Util.isRelativePath;
 import static hudson.util.FormValidation.validateRequired;
 
 /**
@@ -77,6 +78,9 @@ public class Disk implements Describable<Disk> {
         }
 
         public FormValidation doCheckPhysicalPathOnDisk(@QueryParameter String value) {
+            if (!isRelativePath(value)) {
+                return FormValidation.error("Must be a relative path");
+            }
             return validateRequired(value);
         }
 
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
@@ -1,19 +1,24 @@
 package org.jenkinsci.plugins.ewm.steps;
 
 import com.google.inject.Inject;
+import hudson.AbortException;
 import hudson.FilePath;
 import hudson.model.Computer;
 import hudson.model.Node;
-import hudson.model.Run;
+import hudson.model.TaskListener;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.NodePropertyDescriptor;
 import hudson.slaves.WorkspaceList;
 import hudson.util.DescribableList;
 import org.jenkinsci.plugins.ewm.definitions.Template;
+import org.jenkinsci.plugins.ewm.nodes.DiskNode;
 import org.jenkinsci.plugins.ewm.nodes.ExternalWorkspaceProperty;
+import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
 import org.jenkinsci.plugins.workflow.steps.*;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.io.IOException;
 import java.util.List;
 
 /**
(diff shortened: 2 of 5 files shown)


oleg-nenashev:
Retriggering the build

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -76,6 +77,9 @@ public FormValidation doCheckName(@QueryParameter String value) {
         }
 
         public FormValidation doCheckPhysicalPathOnDisk(@QueryParameter String value) {
+            if (!isRelativePath(fixEmptyAndTrim(value))) {
```
oleg-nenashev: StringUtils.isBlank() is preferable

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -67,8 +79,104 @@ public boolean start() throws Exception {
...
+    private static Template findTemplate(String diskPoolRefId, String nodeLabelString, List<Template> templates) throws IOException {
+        Template selectedTemplate = null;
+        for (Template template : templates) {
+            String templateLabel = template.getLabel();
+            if (templateLabel != null && nodeLabelString.contains(templateLabel)) {
+                selectedTemplate = template;
+                break;
+            }
+        }
+        if (selectedTemplate != null) {
```
martinda: You could replace this line with an early return `if (selectedTemplate == null) return null`, but no strong opinion here.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -67,8 +79,104 @@ public boolean start() throws Exception {
...
+            String templateDiskPoolRefId = selectedTemplate.getDiskPoolRefId();
+            if (templateDiskPoolRefId == null) {
+                throw new AbortException("The Disk Pool Ref ID was not provided in Jenkins global config");
```
martinda: This is an error about the `selectedTemplate`. It is missing a `diskPoolRefId`. Try to write an error message that pin points what's missing, so the user will know where to fix it. Could a validation method in the descriptor prevent this error condition?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -67,8 +79,104 @@ public boolean start() throws Exception {
...
+            }
+            if (!templateDiskPoolRefId.equals(diskPoolRefId)) {
+                throw new AbortException("The Disk Pool Ref ID defined in Jenkins global config does not match the one allocated by the exwsAllocate step");
```
martinda: Here we have the template that matches the `nodeStringLabel`, but its `diskPoolRefId` does not match the `diskPoolId` value passed by the user in the `exwsAllocate(id:'the-user-disk-pool-id')`, so the plugin throws an exception.

Try to write an error message that helps the user find which template to fix, or which value of disk pool ref Id to change.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -67,8 +79,104 @@ public boolean start() throws Exception {
...
+        ExternalWorkspaceProperty exwsNodeProperty = null;
+        for (NodeProperty<?> nodeProperty : nodeProperties) {
+            if (nodeProperty instanceof ExternalWorkspaceProperty) {
+                exwsNodeProperty = (ExternalWorkspaceProperty) nodeProperty;
+                break;
+            }
+        }
+        if (exwsNodeProperty == null) {
+            throw new AbortException("There is no External Workspace config defined in Node config");
+        }
```
martinda: I think we forgot a use case where all the nodes have the same mounts as the master node (e.g. in a homogeneous network). Consequently, an exception for this condition here may be too strong. I think we need a second fallback position: the global config. So the search order would be: template, node, global. If after looking up the global config nothing is found, then throw an exception. Perhaps there should also be a short message in the console indicating the search phase being run (so up to three messages). WDYT?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: >  Could a validation method in the descriptor prevent this error condition?

There is a validation error for this, but that error only shows a red sign with `Required` text.
If the user leaves that field empty, he is allowed to submit the form. I've seen this similar behavior for other configs, not only for this plugin.
If you remind me, I'll show you what I mean in our next meeting.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: I don't know yet what to say about having a second fallback. I am trying to think from the user perspective. It may get a little bit to complicated for the user to understand how all these fallbacks are working.
I have created a Jira issue for this, to keep it in mind: https://issues.jenkins-ci.org/browse/JENKINS-35440

I have added logging for the two searches that are currently performed.

martinda:
Thank you Alex!

alexsomai:
This branch was created against `exws-allocate-step` branch, but it should be merged into `master`.
Closing it in favor of #12

PR #8, opened 2016-06-06
[JENKINS-35281] Implement the exwsAllocate step for the downstream job

alexsomai:
The implementation for the exwsAllocate step, the downstream job.
[[JENKINS-35281]](https://issues.jenkins-ci.org/browse/JENKINS-35281)


src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -66,17 +77,41 @@ protected ExternalWorkspace run() throws Exception {
...
+            if (lastStableBuild == null) {
+                throw new AbortException(format("'%s' doesn't have any stable build", upstreamName));
+            }
+            if (!(lastStableBuild instanceof WorkflowRun)) {
+                throw new AbortException(format("Build '%s' is not a Pipeline job. Can't read the run actions", lastStableBuild));
+            }
+
+            FlowExecution flowExecution = ((WorkflowRun) lastStableBuild).getExecution();
+            if (flowExecution == null) {
+                throw new Exception("Upstream flow execution is null");
```
martinda: This is my ignorance asking: why `Exception` here, and `AbortException` in the other ones? Maybe this is temporary until you create the custom exception?

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateAction.java
```diff
@@ -0,0 +1,46 @@
+package org.jenkinsci.plugins.ewm.actions;
+
+import hudson.model.Action;
+import org.jenkinsci.plugins.ewm.Messages;
+import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+/**
+ * {@link Action} implementation for the {@link org.jenkinsci.plugins.ewm.steps.ExwsAllocateStep}.
+ * Stores the allocated {@link ExternalWorkspace}.
```
martinda: I think the doc should say, here or somewhere else, why we store the ext workspace, e.g. for later use in downstream jobs, for removal by another job, etc.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -91,4 +126,42 @@ private String computePathOnDisk(String physicalPathOnDisk) {
...
     }
+
+    /**
+     * Find the upstream job that matches the given upstream name.
+     *
+     * @param upstreamJob the upstream job name
+     * @return the upstream job with the given name if any, {@code null} otherwise
+     */
+    @CheckForNull
+    private Item findUpstreamJob(String upstreamJob) {
```
alexsomai: This is the way how finding the upstream job works:
- the upstream job must be in the same Folder with the downstream job
- the user must pass only the name of the upstream job

I'm not quite sure how finding of the upstream job should work.
@oleg-nenashev you've mentioned about passing the full name when calling `exwsAllocate upstream: 'upstream-full-name'`. In this case, I would have to use `Jenkins.getInstance().getItemByFullName(upstreamJob)`? And this means that the upstream and downstream jobs don't have to be in the same Folder?

I'm saying upstream/downstream in the external workspace manager plugin context, and not how they are defined in Jenkins.

alexsomai:
This PR is reviewable
CC @oleg-nenashev @martinda

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: Mainly I'm using `AbortException` when the build should fail because the user didn't configured the plugin correctly (e.g. missing some parameters)
In this case I've used `Exception` because I don't see what actions the user can take if `flowExecution == null`.

The doc for `getExecution()` says:
>/**
     * Gets the associated execution state.
     * @return non-null after the flow has started, even after finished (but may be null temporarily when about to start, or if starting failed)
     */

I see this as a corner case scenario, very hard (or almost impossible) to reproduce.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
martinda: I think the more general approach is better: the upstream/downstream jobs may not be in the same folder. Oleg said something about relative job names and absolute job names, and complications with computing jobs names (maybe he was thinking about the "job" string being inserted between folder names and job names - but he did not say this explicitly so he may have meant something else).

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateAction.java
oleg-nenashev: Maybe. Not so important for the prototyping stage, when the things may change. Detailed documentation may be a burden at this point

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateAction.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author Alexandru Somai
+ */
+public class ExwsAllocateAction implements Action {
```
oleg-nenashev: I would expect this action to be stored on the top level (with RunAction2 interface).
Otherwise data users will have to search for the data in the build graph, which may require extra resources

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateAction.java
```diff
@@ -0,0 +1,46 @@
...
+        return parent;
+    }
+
+    public ExternalWorkspace getExternalWorkspace() {
+        return externalWorkspace;
+    }
+
+    @Override
+    public String getIconFileName() {
+        return "folder.png";
```
oleg-nenashev: Maybe add TODO for changing it?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
oleg-nenashev: I referenced https://github.com/jenkinsci/promoted-builds-plugin/pull/62, which describes the implementation.
IMHO for a PoC it's better to just use a full name. Maybe I'll create a new API plugin in the short-term

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -81,9 +82,14 @@ protected ExternalWorkspace run() throws Exception {
         } else {
             // this is the downstream job
 
-            Item upstreamJob = findUpstreamJob(upstreamName);
+            if (step.getDiskPoolId() != null) {
+                listener.getLogger().println("WARNING: Both 'upstream' and 'diskPoolId' parameters were provided. " +
+                        "The 'diskPoolId' parameter will be ignored. The step will allocate the workspace used by the upstream job.");
```
alexsomai: @oleg-nenashev @martinda WDYT about this warning message? Is it understandable for the user?

martinda: Yes, it is a good error message.

src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper/index.jelly
```diff
@@ -0,0 +1,11 @@
...
+<st:compress xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout">
+    <l:layout title="${it.displayName}" norefresh="true">
+        <st:include it="${it.parent}" page="sidepanel.jelly"/>
+        <l:main-panel>
+            <j:forEach var="exwsAction" items="${it.allActions}">
+                <st:include it="${exwsAction}" page="index.jelly"/>
+            </j:forEach>
+        </l:main-panel>
+    </l:layout>
+</st:compress>
```
martinda: In one of these jelly files, does it make sense to display to the user, that when multiple allocated workspace have been allocated, only the first one will be used by downstream jobs? Is this a good place to remind the user of this information?

martinda:
From my point of view, I am good with this pull request. Thank you Alex for you work on both these recent pull requests.

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A {@link RunAction2} that acts as a wrapper to hold multiple {@link ExwsAllocateActionImpl}s.
+ *
+ * @author Alexandru Somai
+ */
+public class ExwsAllocateActionWrapper extends Actionable implements RunAction2 {
+
+    private Run parent;
+
+    public Run getParent() {
```
oleg-nenashev: Is it really required? I doubt

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+import hudson.model.Run;
+import jenkins.model.RunAction2;
+import org.jenkinsci.plugins.ewm.Messages;
+
+/**
+ * A {@link RunAction2} that acts as a wrapper to hold multiple {@link ExwsAllocateActionImpl}s.
+ *
+ * @author Alexandru Somai
+ */
+public class ExwsAllocateActionWrapper extends Actionable implements RunAction2 {
```
oleg-nenashev: Ohh, Actionable action is something new :)
It may work, but I doubt it's a right approach until this wrapper provides a kind of API for accessing workspaces.
BTW I would expect this action to store a hashset (better performance than class-indexed array of ExwsAllocateActionImpl).

oleg-nenashev:
I would kindly vote against Action of Actions. You may have to access this data frequently, and such implementation may complicate search operations

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
alexsomai: I knew that having an Actionable Action is a bad idea :) But I don't know what I was thinking when I implemented that.

alexsomai:
I do agree with you. I don't like either having an Action of Actions.

src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper/index.jelly
alexsomai: Yes, this is a good idea.

oleg-nenashev: BTW I doubt that st:compress is really required here. The page is pretty small. Not important

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
alexsomai: I need this to include the sidepanel in the jelly file.
`<st:include it="${it.parent}" page="sidepanel.jelly"/>`
Otherwsie is not working.
Is it another way to do this?

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
alexsomai: If I use a `HashSet`, the set does not guarantee the order of the items.

Use case:  If the user calls `exwsAllocate` step multiple times, the set will contain multiple items. If I have a downstream job that uses this, it is stated that it should use the first registered `exwsAllocate` action. So, if a set is used, I don't have the guarantee which will be the first inserted item.
Therefore I would rather go for a `List`. Maybe a `LinkedList` implementation?

oleg-nenashev: Maybe TreeSet them or Sorted map. No strong opinion

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
oleg-nenashev: No, it's fine then. Just add <code>@Restricted(NoExternalUse)</code> to prevent the wide exposure

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionWrapper.java
oleg-nenashev: Decision from the meeting: Keep it as list now, consider performance tweaks later

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -66,17 +76,51 @@ protected ExternalWorkspace run() throws Exception {
...
+            // this list always contains at least one element
+            exws = exwsAllocateActions.iterator().next().getExternalWorkspace();
         }
+
+        ExwsAllocateActionWrapper exwsActionWrapper = run.getAction(ExwsAllocateActionWrapper.class);
+        if (exwsActionWrapper == null) {
+            exwsActionWrapper = new ExwsAllocateActionWrapper();
+            run.addAction(exwsActionWrapper);
+        }
+        exwsActionWrapper.addAction(new ExwsAllocateActionImpl(exws));
```
oleg-nenashev: Just keep in mind you will need to invoke save() to persist changes

alexsomai:
I've pushed the changes made after the discussions that we had.

src/main/java/org/jenkinsci/plugins/ewm/steps/model/ExternalWorkspace.java
```diff
@@ -2,27 +2,35 @@
...
 
     private static final long serialVersionUID = 1L;
 
+    private final UUID id;
     private final String diskPoolId;
     private final String diskId;
     private final String pathOnDisk;
 
     public ExternalWorkspace(@Nonnull String diskPoolId, @Nonnull String diskId, @Nonnull String pathOnDisk) {
+        this.id = UUID.randomUUID();
```
alexsomai: I don't think I can use https://github.com/jenkinsci/unique-id-plugin for generating workspace unique id.
From my understanding, if I use `IdStore.makeId(Object o)` it stores a generated id into a file. But I don't think I need that kind of functionality.

oleg-nenashev: The current approach is fine for PoC, but maybe it's better to store the String instead of UUID in the field

alexsomai: Yeap, good point

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -94,28 +90,31 @@ protected ExternalWorkspace run() throws Exception {
...
             }
 
-            ExwsAllocateActionWrapper exwsActionWrapper = lastStableBuild.getAction(ExwsAllocateActionWrapper.class);
-            if (exwsActionWrapper == null) {
-                String message = format("The Jenkins job '%s' does not have registered any 'External Workspace Allocate' action. Did you run exwsAllocate step in the upstream job?", upstreamName);
+            ExwsAllocateActionImpl allocateAction = lastStableBuild.getAction(ExwsAllocateActionImpl.class);
+            if (allocateAction == null) {
+                String message = format("This downstream job has to allocate the same external workspace used by the upstream job. " +
+                        "To do so, the exwsAllocate step must be called in the upstream job. " +
+                        "Please call the exwsAllocate step in the upstream job, because the build '%s' doesn't have such calls.", lastStableBuild);
```
martinda: What do you think about this:
"No stable build upstream workspace could be found. The upstream job must have at least one stable build with a call to the exwsAllocate step in order to have a workspace usable by this job."

alexsomai: It's better, but I have one mention: if the upstream job doesn't have any stable build, the user will get the error message from line [90](https://github.com/jenkinsci/external-workspace-manager-plugin/pull/8/files#diff-fe1322463bd4bb50e34cee4801cbfd3eR90). Therefore, the first sentence: "No stable build upstream workspace could be found" may be a little bit redundant.

So, IMO it would be enough to say: "The upstream job must have at least one stable build with a call to the exwsAllocate step in order to have a workspace usable by this job."

WDYT?

src/main/java/org/jenkinsci/plugins/ewm/steps/model/ExternalWorkspace.java
```diff
@@ -2,27 +2,35 @@
...
  * POJO used to pass fields from one step to another.
  *
  * @author Alexandru Somai
  */
-public final class ExternalWorkspace implements Serializable {
+public class ExternalWorkspace implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    private final String id;
```
oleg-nenashev: NIT for now, but in after 1.0 such changes will require readResolve() for data migration

alexsomai:
This PR had to be created against `master` branch. Closing in favor of #13.

PR #10, opened 2016-06-07
[JENKINS-35192] [JENKINS-35244] [JENKINS-35281] Unit tests

alexsomai:
Write unit tests for the features implemented so far.

[[JENKINS-35192]](https://issues.jenkins-ci.org/browse/JENKINS-35192) Unit tests for the exwsAllocate step
[[JENKINS-35244]](https://issues.jenkins-ci.org/browse/JENKINS-35244) Unit tests for the exws step
[[JENKINS-35281]](https://issues.jenkins-ci.org/browse/JENKINS-35281) Unit tests for the exwsAllocate step - downstream job

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -57,13 +57,13 @@ public class ExwsAllocateExecution extends AbstractSynchronousNonBlockingStepExe
             Disk disk = allocationStrategy.allocateDisk();
 
             String physicalPathOnDisk = disk.getPhysicalPathOnDisk();
-            if (physicalPathOnDisk == null) {
-                String message = format("Physical path on disk was not provided in the Jenkins global config for the Disk Pool ID '%s'", diskPoolId);
-                throw new AbortException(message);
-            }
             String diskId = disk.getDiskId();
             if (diskId == null) {
                 String message = format("Disk ID was not provided in the Jenkins global config for the Disk Pool ID '%s'", diskPoolId);
+                throw new AbortException(message);
+            }
+            if (physicalPathOnDisk == null) {
+                String message = format("Physical path on disk was not provided in the Jenkins global config for Disk ID: '%s', within Disk Pool ID '%s'", diskId, diskPoolId);
                 throw new AbortException(message);
             }
             if (!isRelativePath(physicalPathOnDisk)) {
--- a/src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
@@ -1,0 +1,205 @@
+package org.jenkinsci.plugins.ewm.steps;
+
+import hudson.model.FreeStyleBuild;
+import hudson.model.queue.QueueTaskFuture;
+import hudson.util.ReflectionUtils;
+import org.jenkinsci.plugins.ewm.definitions.Disk;
+import org.jenkinsci.plugins.ewm.definitions.DiskPool;
+import org.jenkinsci.plugins.ewm.steps.ExwsAllocateStep.DescriptorImpl;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static hudson.model.Result.FAILURE;
+import static java.lang.String.format;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Unit tests for {@link ExwsAllocateStep}.
+ *
+ * @author Alexandru Somai
+ */
+public class ExwsAllocateStepTest {
+
+    private static final String DISK_POOL_ID = "disk-pool-1";
+    private static final String DISK_ID = "disk1";
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    private WorkflowRun upstreamRun;
+    private WorkflowRun downstreamRun;
+
+    /* ##### Tests for the upstream Job ###### */
+
+    @Test
+    public void missingDiskPoolId() throws Exception {
+        createUpstreamJobAndRun(" ");
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains("Disk Pool ID was not provided as step parameter", upstreamRun);
+    }
+
+    @Test
+    public void wrongDiskPoolId() throws Exception {
+        setupDiskPool();
+        createUpstreamJobAndRun(DISK_POOL_ID + "random");
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("No Disk Pool ID matching '%s' was found in the global config", DISK_POOL_ID + "random"), upstreamRun);
+    }
+
+    @Test
+    public void emptyDisks() throws Exception {
+        setupDiskPool();
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("No Disks were defined in the global config for Disk Pool ID '%s'", DISK_POOL_ID), upstreamRun);
+    }
+
+    @Test
+    public void missingDiskId() throws Exception {
+        setupDiskPool(new Disk("", "name", "mount", "path"));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Disk ID was not provided in the Jenkins global config for the Disk Pool ID '%s'", DISK_POOL_ID), upstreamRun);
+    }
+
+    @Test
+    public void missingMasterMountPoint() throws Exception {
+        setupDiskPool(new Disk(DISK_ID, "name", "", "path"));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Mounting point from Master to the disk is not defined for Disk ID '%s', from Disk Pool ID '%s'", DISK_ID, DISK_POOL_ID), upstreamRun);
+    }
+
+    @Test
+    public void missingPhysicalPathOnDisk() throws Exception {
+        setupDiskPool(new Disk(DISK_ID, "name", "mount", ""));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Physical path on disk was not provided in the Jenkins global config for Disk ID: '%s', within Disk Pool ID '%s'", DISK_ID, DISK_POOL_ID), upstreamRun);
+    }
+
+    @Test
+    public void physicalPathOnDiskNotRelative() throws Exception {
+        setupDiskPool(new Disk(DISK_ID, "name", "mount", "/path"));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Physical path on disk defined for Disk ID '%s', within Disk Pool ID '%s' must be a relative path", DISK_ID, DISK_POOL_ID), upstreamRun);
+    }
+
+    @Test
+    public void successfullyAllocateWorkspace() throws Exception {
+        Disk disk = new Disk(DISK_ID, "name", "mount", "path");
+        setupDiskPool(disk);
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatusSuccess(upstreamRun);
+        j.assertLogContains(format("Selected Disk ID '%s' from the Disk Pool ID '%s'", DISK_ID, DISK_POOL_ID), upstreamRun);
+        j.assertLogContains(format("The path on Disk is: %s/%s/%d", disk.getPhysicalPathOnDisk(), upstreamRun.getParent().getFullDisplayName(), upstreamRun.getNumber()), upstreamRun);
+    }
+
+    /* ##### Tests for the downstream Job ###### */
+
+    @Test
+    public void downstreamJobWithInvalidUpstreamName() throws Exception {
+        String upstreamName = "random-upstream-name";
+        createDownstreamJobAndRun(upstreamName);
+
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("There isn't any upstream job associated with '%s'", upstreamName), downstreamRun);
+    }
+
+    @Test
+    public void upstreamJobHasNoStableBuild() throws Exception {
+        createUpstreamJobAndRun("random-disk-pool");
+        String upstreamName = upstreamRun.getParent().getName();
+        createDownstreamJobAndRun(upstreamName);
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("'%s' doesn't have any stable build", upstreamName), downstreamRun);
+    }
+
+    @Test
+    public void upstreamJobHasNoActionRegistered() throws Exception {
+        String jobName = "test";
+        createWorkflowJobAndRun(jobName, "echo 'hello world'");
+        createDownstreamJobAndRun(jobName);
+
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("The Jenkins job '%s' does not have registered any 'External Workspace Allocate' action", jobName), downstreamRun);
+    }
+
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -62,15 +62,15 @@ protected ExternalWorkspace run() throws Exception {
...
-            if (physicalPathOnDisk == null) {
-                String message = format("Physical path on disk was not provided in the Jenkins global config for the Disk Pool ID '%s'", diskPoolId);
-                throw new AbortException(message);
-            }
             String diskId = disk.getDiskId();
             if (diskId == null) {
                 String message = format("Disk ID was not provided in the Jenkins global config for the Disk Pool ID '%s'", diskPoolId);
                 throw new AbortException(message);
             }
+            if (physicalPathOnDisk == null) {
```
oleg-nenashev: Was it a bugfix?

alexsomai: No, I have just changed the order of null checks.
When I was writing the tests, I though it was more appropriate to check firstly if the diskId is null, and the check if the physicalPathOnDisk is null.
Before, I was checking if the physicalPathOnDisk is null, and afterwards if the diskId is null.

alexsomai:
This PR is ready for review.
CC @oleg-nenashev @martinda

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void missingMasterMountPoint() throws Exception {
+        setupDiskPool(new Disk(DISK_ID, "name", "", "path"));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Mounting point from Master to the disk is not defined for Disk ID '%s', from Disk Pool ID '%s'", DISK_ID, DISK_POOL_ID), upstreamRun);
+    }
```
martinda: I think it is okay to have the `missingMasterMountPoint` test here for now, but when there is more than one allocation strategy, it may be better to place strategy testing in a test file of its own. At this point, no change is needed IMO.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,205 @@
...
+        Disk disk = new Disk(DISK_ID, "name", "mount", "path");
+        setupDiskPool(disk);
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatusSuccess(upstreamRun);
+        j.assertLogContains(format("Selected Disk ID '%s' from the Disk Pool ID '%s'", DISK_ID, DISK_POOL_ID), upstreamRun);
+        j.assertLogContains(format("The path on Disk is: %s/%s/%d", disk.getPhysicalPathOnDisk(), upstreamRun.getParent().getFullDisplayName(), upstreamRun.getNumber()), upstreamRun);
+    }
+
+    /* ##### Tests for the downstream Job ###### */
```
martinda: Not important: my personal preference would have been to split the upstream and the downstream tests into two test files, but this is not a priority for now. The tests do run very fast on my computer, so no issue in terms of execution time. No change needed at this time.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void upstreamJobHasNoActionRegistered() throws Exception {
+        String jobName = "test";
+        createWorkflowJobAndRun(jobName, "echo 'hello world'");
+        createDownstreamJobAndRun(jobName);
+
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("The Jenkins job '%s' does not have registered any 'External Workspace Allocate' action", jobName), downstreamRun);
```
martinda: I think better grammar would be "... hasn't register any 'External ...", but I think the error message could be made more clear to the user.

When I write error messages in my code, I try to tell the user how they got in trouble in the first place, and how they can fix it. When I read this code, I ask myself: the lack of "run action" is a symptom of what? We want to make sure the novice user does not go on a tangent looking for ways to add a "run action", when all they have to do is call `exwsAllocate`. Ok, so let's park this idea and look at some corner cases.

I think there is a corner case where the user has a build history like this:
2. upstream calls `exwsAllocate` but fails **before** it registers the action.
1. upstream does not call `exwsAllocate` and is stable (green).

And another corner case, almost identical to the case above:
2. upstream calls `exwsAllocate` but fails **after** it registers the action
1. upstream does not call `exwsAllocate` and is stable (green)

We want the user to have a build that is stable and that calls `exwsAllocate`, so it has the "action" the code is looking for. Try to write an error message that guides the user in this direction.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void upstreamBuildIsNotPipelineJob() throws Exception {
+        String jobName = "name";
+        FreeStyleBuild build = j.createFreeStyleProject(jobName).scheduleBuild2(0).get();
+        createDownstreamJobAndRun(jobName);
+
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("Build '%s' is not a Pipeline job. Can't read the run actions", build), downstreamRun);
```
martinda: Not important: I can't seem to find the code that causes this error message in the plugin code, so I guess it comes from the pipeline code?

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Test
+    public void missingMasterMountPoint() throws Exception {
+        setUpDiskPool(new Disk(DISK_ID_ONE, "name", "", "path"));
+        createUpstreamJobAndRun();
+
+        j.assertBuildStatus(FAILURE, upstreamRun);
+        j.assertLogContains(format("Mounting point from Master to the disk is not defined for Disk ID '%s', from Disk Pool ID '%s'", DISK_ID_ONE, DISK_POOL_ID), upstreamRun);
+    }
```
martinda: I think it is okay to have the `missingMasterMountPoint` test here for now, but when there is more than one allocation strategy, it may be better to place strategy testing in a test file of its own. At this point, no change is needed IMO.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,247 @@
...
+        createUpstreamJobAndRun();
+
+        Disk allocatedDisk = findAllocatedDisk(disk1, disk2);
+
+        j.assertBuildStatusSuccess(upstreamRun);
+        j.assertLogContains(format("Selected Disk ID '%s' from the Disk Pool ID '%s'", allocatedDisk.getDiskId(), DISK_POOL_ID), upstreamRun);
+        j.assertLogContains(format("The path on Disk is: %s/%s/%d", allocatedDisk.getPhysicalPathOnDisk(), upstreamRun.getParent().getFullDisplayName(), upstreamRun.getNumber()), upstreamRun);
+    }
+
+    /* ##### Tests for the downstream Job ###### */
```
martinda: Not important: my personal preference would have been to split the upstream and the downstream tests into two test files, but this is not a priority for now. The tests do run very fast on my computer, so no issue in terms of execution time. No change needed at this time.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Test
+    public void upstreamJobHasNoActionRegistered() throws Exception {
+        String jobName = randomAlphanumeric(10);
+        createWorkflowJobAndRun(jobName, "echo 'hello world'");
+        createDownstreamJobAndRun(jobName);
+
+        j.assertBuildStatus(FAILURE, downstreamRun);
+        j.assertLogContains(format("The Jenkins job '%s' does not have registered any 'External Workspace Allocate' action. Did you run exwsAllocate step in the upstream job?", jobName), downstreamRun);
```
martinda: I think better grammar would be "... hasn't register any 'External ...", but I think the error message could be made more clear to the user.

When I write error messages in my code, I try to tell the user how they got in trouble in the first place, and how they can fix it. When I read this code, I ask myself: the lack of "run action" is a symptom of what? We want to make sure the novice user does not go on a tangent looking for ways to add a "run action", when all they have to do is call `exwsAllocate`. Ok, so let's park this idea and look at some corner cases.

I think there is a corner case where the user has a build history like this:
2. upstream calls `exwsAllocate` but fails **before** it registers the action.
1. upstream does not call `exwsAllocate` and is stable (green).

And another corner case, almost identical to the case above:
2. upstream calls `exwsAllocate` but fails **after** it registers the action
1. upstream does not call `exwsAllocate` and is stable (green)

We want the user to have a build that is stable and that calls `exwsAllocate`, so it has the "action" the code is looking for. Try to write an error message that guides the user in this direction.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -0,0 +1,247 @@
...
+        j.assertBuildStatusSuccess(upstreamRun);
+        j.assertBuildStatusSuccess(downstreamRun);
+        j.assertLogContains("WARNING: Both 'upstream' and 'diskPoolId' parameters were provided. " +
+                "The 'diskPoolId' parameter will be ignored. The step will allocate the workspace used by the upstream job.", downstreamRun);
+        j.assertLogContains(format("Selected Disk ID '%s' from the Disk Pool ID '%s'", DISK_ID_ONE, DISK_POOL_ID), downstreamRun);
+        j.assertLogContains(format("The path on Disk is: %s/%s/%d", disk.getPhysicalPathOnDisk(), upstreamName, upstreamRun.getNumber()), downstreamRun);
+    }
+
+    @Test
+    public void upstreamJobRegisteredMultipleActions() throws Exception {
```
martinda: It is very good that you thought about this case.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsStepTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        j.assertBuildStatusSuccess(run);
+        j.assertLogContains("Searching for disk definitions in the External Workspace Templates from Jenkins global config", run);
+        j.assertLogNotContains("Searching for disk definitions in the Node config", run);
+        j.assertLogContains(format("Running in %s/%s/%s/%d",
+                allocatedDisk.getMasterMountPoint(), allocatedDisk.getPhysicalPathOnDisk(), run.getParent().getName(), run.getNumber()), run);
+        // The text written to file should be printed twice on the console output (when writing and when reading the file)
+        assertThat(countMatches(JenkinsRule.getLog(run), TEXT), is(2));
+    }
+
```
martinda: I can think of one more test: a test with two different workflow jobs using the same workspace. Not so much to prove that it works, but to prevent future modifications to the plugin from breaking it without noticing.

martinda:
From my point of view, just two items need to be addressed, and this pull request is ready.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: I was thinking to add test classes for each allocation strategy when we'll have multiple implementations.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: This is something I'll take into consideration when multiple features will be implemented for the `exwsAllocate` step.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: The current branch version doesn't have this test anymore.
This was a test case for an older implementation: when `exwsAllocate` step was registering the run actions on the `flowNode` object. Now it registers the run action on the `run` object.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: Please let me know if you find future grammar mistakes, since my english has some flaws :)

Regarding the corner cases:
> .2. upstream calls exwsAllocate but fails **before** it registers the action.

and 

> .2. upstream calls exwsAllocate but fails **after** it registers the action

If the build fails, no matter at which point, that build number will not be taken into consideration anyway.

But I can improve this error message for the case when the upstream build is green, but the user didn't call `exwsAllocate` step.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: I was thinking to add test classes for each allocation strategy when we'll have multiple implementations.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: Please let me know if you find future grammar mistakes, since my english has some flaws :)

Regarding the corner cases:

> .2. upstream calls exwsAllocate but fails before it registers the action.

and

> .2. upstream calls exwsAllocate but fails after it registers the action
If the build fails, no matter at which point, that build number will not be taken into consideration anyway.

But I can improve this error message for the case when the upstream build is green, but the user didn't call exwsAllocate step.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
alexsomai: This is something I'll take into consideration when multiple features will be implemented for the exwsAllocate step.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsStepTest.java
alexsomai: Yes, good point. I'll add that.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
```diff
@@ -159,11 +159,13 @@ public void upstreamJobHasNoStableBuild() throws Exception {
...
         String jobName = randomAlphanumeric(10);
-        createWorkflowJobAndRun(jobName, "echo 'hello world'");
+        upstreamRun = createWorkflowJobAndRun(jobName, "echo 'hello world'");
         createDownstreamJobAndRun(jobName);
 
         j.assertBuildStatus(FAILURE, downstreamRun);
-        j.assertLogContains(format("The Jenkins job '%s' does not have registered any 'External Workspace Allocate' action. Did you run exwsAllocate step in the upstream job?", jobName), downstreamRun);
+        j.assertLogContains(format("This downstream job has to allocate the same external workspace used by the upstream job. " +
+                "To do so, the exwsAllocate step must be called in the upstream job. " +
+                "Please call the exwsAllocate step in the upstream job, because the build '%s' doesn't have such calls.", upstreamRun), downstreamRun);
```
alexsomai: I have improved this error message. The actual change is in #8, this is only the test associated with it.

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsStepTest.java
```diff
@@ -215,6 +215,28 @@ public void sharedWorkspaceBetweenTwoDifferentNodesWithTemplate() throws Excepti
         assertThat(countMatches(JenkinsRule.getLog(run), TEXT), is(2));
     }
 
+    @Test
+    public void twoDifferentJobsUsingTheSameWorkspace() throws Exception {
```
alexsomai: I have added one more test case that @martinda suggested.

oleg-nenashev:
Looks good to me. I would merge it in order to get some testing in the new PRs.
Maybe there is some minor changes and extra tests required, but should work fine for the current phase

oleg-nenashev:
BTW a good practice is to summarize changes and to add issue hyperlinks to the PR summary

src/test/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStepTest.java
martinda: I commented on the actual change in PR #8 . I find it difficult to come up with a good error message for this error condition, but please see my comment, I think I have nailed it.

alexsomai:
This PR had to be created against `master` branch. Closing in favor of #14.

PR #11, opened 2016-06-09
Add the build status icon to the README file

alexsomai:


oleg-nenashev:
This approach has some downsides, because it always shows the latest build. So if somebody creates a failing PR, this status will become red on the main page.

Since you have a CI, there is no real need to show this status:
* master builds will be always fine excepting cases of INFRA issues or instable tests
* for pull requests statuses will be displayed directly in their pages

alexsomai:
I though that it will show only the master build. 
But indeed, there is no need for this, if it shows the status for each PR. In this case it will be more confusing.
I'll close this PR.

oleg-nenashev:
I hope we rework building approach and make it useful.
E.g. migration to Pipeline Multibranch may address the case

alexsomai:
I've updated the build status to point at [ci.jenkins.io](https://ci.jenkins.io/job/Plugins/job/external-workspace-manager-plugin/job/master/), master branch.
Of course that most of the time the build on master branch will be passing, but maybe in this way, others can see how the Jenkinsfile works.

oleg-nenashev:
👍

PR #17, opened 2016-06-30
[JENKINS-35653] Enhance help files

alexsomai:
Related to issue [[JENKINS-35653]](https://issues.jenkins-ci.org/browse/JENKINS-35653)

Summary of this pull request:
 - Enhance help files
 - Make _Physical path on disk_ parameter optional

CC @oleg-nenashev @martinda


src/main/resources/org/jenkinsci/plugins/ewm/definitions/Template/help-label.html
```diff
@@ -1,3 +1,17 @@
...
+    </p>
+    <p>
+        <b>Note:</b> If you specify the <i>Disk ID</i> and the <i>Local root path</i> in the
+        <i>External Workspace Templates</i> for a specific Node label, then these parameters will be taken
+        into consideration by the plugin.
+        If the parameters are also provided in a Node config having the given label, then the parameters from the
+        Node config will be ignored.
+        But, if these two parameters are not provided in the <i>External Workspace Templates</i>,
+        they must be mandatory provided in each Node config.
+    </p>
```
martinda: What do you think of this shorter documentation:

When the actual Node label (in the Node configuration) matches the *Node label* in this *External Workspace Template*, then the disks in this template are used by the plugin.

Otherwise you must configure a disk in the Node configuration page under *Node Properties*, *External Workspace*.

src/main/resources/org/jenkinsci/plugins/ewm/nodes/DiskNode/help-localRootPath.html
```diff
@@ -1,3 +1,10 @@
 <div>
-    The mounting point from this node to the actual disk.
+    <p>
+        The mounting point (via NFS, Samba, etc.) from Jenkins Node to the actual disk.
+    </p>
+    <p>
+        This parameter is used to compute the complete workspace path within a Node.
+        When the complete workspace path is computed, it follows the following pattern:
+        <i>/localRootPath/physicalPathOnDisk/$JOB_NAME/$BUILD_NUMBER</i>
```
martinda: The *physicalPathOnDisk* is defined in the Jenkins global config, *External Workspace Definitions* section.

alexsomai:
In the issue was also mentioned to have a more generic help file for the *External Workspace Template* and for the *External Workspace* Node Property, but I didn't find a way to add that. I don't think it's possible.

martinda:
Yes, there is a way to provide a help file at the top level of the plugin. The http-request plugin does that [here](https://github.com/jenkinsci/http-request-plugin/blob/master/src/main/resources/jenkins/plugins/http_request/HttpRequestStep/help.html). I think you simply have to set the name of the help file to `help.html` under `src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsStep/`.

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -32,7 +31,7 @@ public Disk(String diskId, String name, String masterMountPoint, String physical
         this.diskId = fixEmptyAndTrim(diskId);
         this.name = fixEmptyAndTrim(name);
         this.masterMountPoint = fixEmptyAndTrim(masterMountPoint);
-        this.physicalPathOnDisk = fixEmptyAndTrim(physicalPathOnDisk);
+        this.physicalPathOnDisk = fixNull(physicalPathOnDisk).trim();
```
oleg-nenashev: It's unrelated to help files, I'd guess. Maybe was better to keep null in fields and process it in the logic

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -47,15 +46,15 @@ public String getDiskId() {
...
 
     @CheckForNull
     public String getMasterMountPoint() {
         return masterMountPoint;
     }
 
-    @CheckForNull
+    @Nonnull
     public String getPhysicalPathOnDisk() {
         return physicalPathOnDisk;
```
oleg-nenashev: This method may still return null if somebody updates from alpha-1. The best approach would be to add readResolve() method (see https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility). Not important for the current project stage

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -11,8 +11,7 @@
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
-import static hudson.Util.fixEmptyAndTrim;
-import static hudson.Util.isRelativePath;
+import static hudson.Util.*;
```
oleg-nenashev: It's better to avoid doing it in the production code, because dependency updates may lead to the conflicts in the code

src/main/resources/org/jenkinsci/plugins/ewm/definitions/Disk/config.jelly
```diff
@@ -3,7 +3,7 @@
     <f:entry title="${%Disk ID}" field="diskId">
         <f:textbox/>
     </f:entry>
-    <f:entry title="${%Name}" field="name">
+    <f:entry title="${%Display Name}" field="name">
```
oleg-nenashev: maybe makes sense to rename the field as well

src/main/resources/org/jenkinsci/plugins/ewm/definitions/Template/help-label.html
```diff
@@ -1,3 +1,10 @@
 <div>
-    Node's label. Will use this template configuration for all the nodes matching the given label.
+    <p>
+        When the actual Node label (in the Node configuration) matches the <i>Node label</i> in this <i>External
```
oleg-nenashev: Actually, is it a single label? Does it support label expressions? E.g. "linux && 64bit"

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
alexsomai: You are right. I'll revert this and have the logic in the step execution.

src/main/resources/org/jenkinsci/plugins/ewm/definitions/Template/help-label.html
alexsomai: Yes, currently it's just a single label. I didn't implement the feature for label expressions. 
I thought that in a first phase would be ok like that. And afterwards I would make it to support label expressions.

src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
alexsomai: I would avoid adding *readResolve()* method at this stage. It's still an alpha version, and the users (if there are any) are aware that things may change and the plugin may break.

src/main/java/org/jenkinsci/plugins/ewm/definitions/DiskPool.java
```diff
@@ -25,14 +25,14 @@
...
 
     private final String diskPoolId;
-    private final String name;
+    private final String displayName;
     private final String description;
     private final List<Disk> disks;
 
     @DataBoundConstructor
-    public DiskPool(String diskPoolId, String name, String description, List<Disk> disks) {
+    public DiskPool(String diskPoolId, String displayName, String description, List<Disk> disks) {
```
oleg-nenashev: Not for this PR. It would be great to prevent null IDs from the API complexity PoV

oleg-nenashev:
👍

martinda:
+1

src/main/java/org/jenkinsci/plugins/ewm/definitions/DiskPool.java
alexsomai: I've create this issue to keep track of what you've mentioned: https://issues.jenkins-ci.org/browse/JENKINS-36761

PR #19, opened 2016-07-04
[JENKINS-36417] PoC: Integrate BuildSelector extension point

alexsomai:
Related to issue [[JENKINS-36417]](https://issues.jenkins-ci.org/browse/JENKINS-36417)

Summary of this pull request:
 - Allow the user to specify multiple build selection strategies by integrating the BuildSelector extension point from Copy Artifact plugin.

A few possible selection strategies:

1. Specific build number from the upstream job
   ```groovy
   def exwsWorkspace = exwsAllocate upstream: 'upstream-full-name',
selector: [$class: 'SpecificRunSelector', buildNumber: '20']
   ```

1. Last stable build from the upstream job
   ```groovy
   def exwsWorkspace = exwsAllocate upstream: 'upstream-full-name', 
selector: [$class: 'StatusRunSelector', buildStatus: 'Stable']
   ```

1. The upstream build that triggered the downstream
 - Upstream job

   ```groovy
   // ...
   build ('full-downstream-name')
   ``` 
 - Downstream job

   ```groovy
   def exwsWorkspace = exwsAllocate upstream: 'upstream-full-name', 
selector: [$class: 'TriggeringRunSelector']
   ```

CC @martinda @oleg-nenashev

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -18,6 +18,8 @@ import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
 import org.jenkinsci.plugins.ewm.strategies.DiskAllocationStrategy;
 import org.jenkinsci.plugins.ewm.strategies.MostUsableSpaceStrategy;
 import org.jenkinsci.plugins.runselector.context.RunSelectorPickContext;
+import org.jenkinsci.plugins.runselector.filters.NoRunFilter;
+import org.jenkinsci.plugins.runselector.filters.RunFilter;
 import org.jenkinsci.plugins.runselector.selectors.RunSelector;
 import org.jenkinsci.plugins.runselector.selectors.StatusRunSelector;
 import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
@@ -110,7 +112,8 @@ public class ExwsAllocateExecution extends AbstractSynchronousNonBlockingStepExe
 
             String jobName = envVars.expand(upstreamName);
             context.setProjectName(jobName);
-            context.setRunFilter(step.getRunFilter());
+            RunFilter runFilter = step.getRunFilter() != null ? step.getRunFilter() : new NoRunFilter();
+            context.setRunFilter(runFilter);
 
             Run<?, ?> upstreamBuild = selector.pickBuildToCopyFrom(upstreamJob, context);
 
--- a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
@@ -10,10 +10,6 @@ import hudson.model.ParameterValue;
 import hudson.model.ParametersAction;
 import hudson.model.Run;
 import hudson.model.TaskListener;
-import hudson.plugins.copyartifact.BuildFilter;
-import hudson.plugins.copyartifact.BuildSelector;
-import hudson.plugins.copyartifact.ParametersBuildFilter;
-import hudson.plugins.copyartifact.StatusBuildSelector;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.ewm.actions.ExwsAllocateActionImpl;
 import org.jenkinsci.plugins.ewm.definitions.Disk;
@@ -21,6 +17,9 @@ import org.jenkinsci.plugins.ewm.definitions.DiskPool;
 import org.jenkinsci.plugins.ewm.steps.model.ExternalWorkspace;
 import org.jenkinsci.plugins.ewm.strategies.DiskAllocationStrategy;
 import org.jenkinsci.plugins.ewm.strategies.MostUsableSpaceStrategy;
+import org.jenkinsci.plugins.runselector.context.RunSelectorPickContext;
+import org.jenkinsci.plugins.runselector.selectors.RunSelector;
+import org.jenkinsci.plugins.runselector.selectors.StatusRunSelector;
 import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
 import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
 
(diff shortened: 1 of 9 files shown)


martinda:
Interestingly small for such a useful feature. This proof of concept looks good to me (with my limited experience in coding for Jenkins).

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -76,21 +85,32 @@ protected ExternalWorkspace run() throws Exception {
...
-            Item upstreamJob = Jenkins.getActiveInstance().getItemByFullName(upstreamName);
+            Job<?, ?> upstreamJob = Jenkins.getActiveInstance().getItemByFullName(upstreamName, Job.class);
             if (upstreamJob == null) {
                 throw new AbortException(format("Can't find any upstream Jenkins job by the full name '%s'. Are you sure that this is the full project name?", upstreamName));
             }
-            Run lastStableBuild = ((Job) upstreamJob).getLastStableBuild();
-            if (lastStableBuild == null) {
-                throw new AbortException(format("'%s' doesn't have any stable build", upstreamName));
+
+            EnvVars envVars = getEnvVars();
```
oleg-nenashev: I would double-check that EnvVars work fine for Pipeline

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -76,21 +85,32 @@ protected ExternalWorkspace run() throws Exception {
...
+
+            Run<?, ?> upstreamBuild = selector.getBuild(upstreamJob, envVars, buildFilter, run);
+            if (upstreamBuild == null) {
+                throw new AbortException(format("Unable to find a build within upstream job '%s'", upstreamName));
             }
 
-            ExwsAllocateActionImpl allocateAction = lastStableBuild.getAction(ExwsAllocateActionImpl.class);
+            ExwsAllocateActionImpl allocateAction = upstreamBuild.getAction(ExwsAllocateActionImpl.class);
             if (allocateAction == null) {
                 String message = format("The upstream job '%s' must have at least one stable build with a call to the " +
```
oleg-nenashev: This message is obsolete && misleading now

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -76,21 +85,32 @@ protected ExternalWorkspace run() throws Exception {
...
-                throw new AbortException(format("'%s' doesn't have any stable build", upstreamName));
+
+            EnvVars envVars = getEnvVars();
+            String parameters = step.getParameters();
+            BuildFilter buildFilter = parameters != null ? new ParametersBuildFilter(envVars.expand(parameters)) : new BuildFilter();
+
+            Run<?, ?> upstreamBuild = selector.getBuild(upstreamJob, envVars, buildFilter, run);
+            if (upstreamBuild == null) {
+                throw new AbortException(format("Unable to find a build within upstream job '%s'", upstreamName));
             }
```
oleg-nenashev: Maybe makes sense to add a log message about the selected build

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -122,6 +142,28 @@ protected ExternalWorkspace run() throws Exception {
     }
 
     /**
+     * Gets the environment variables based on the current run.
+     *
+     * @return the {@link EnvVars} for the current run
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    private EnvVars getEnvVars() throws IOException, InterruptedException {
```
oleg-nenashev: 🐛 This code is misleading, because EnvVars are not only about Build Parameters.
IMHO you should use <code>Run#getEnvironment(TaskListener</code>) instead of this home-made method

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -48,6 +51,26 @@ public void setUpstream(String upstream) {
...
+        this.selector = selector;
+    }
+
+    @CheckForNull
+    public String getParameters() {
+        return parameters;
+    }
+
+    @DataBoundSetter
+    public void setParameters(String parameters) {
```
oleg-nenashev: Would it be possible to put these parematerFilters into the Selector directly? The current API location is not very good  IMHO

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -27,6 +28,8 @@
 
     private final String diskPoolId;
     private String upstream;
+    private BuildSelector selector;
```
oleg-nenashev: Good practice: Annotate fields as well in order to avoid misusages in the code

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
alexsomai: Mostly what I did was to copy/paste from the Copy Artifact plugin. It had some additional comments in there. I quote:
> ```java
// Abstract#getEnvironment(TaskListener) put build parameters to
// environments, but Run#getEnvironment(TaskListener) doesn't.
// That means we can't retrieve build parameters from WorkflowRun
// as it is a subclass of Run, not of AbstractBuild.
// We need expand build parameters manually.
// See JENKINS-26694, JENKINS-30357 for details.
for(ParametersAction pa: build.getActions(ParametersAction.class)) {
    // We have to extract parameters manally as ParametersAction#buildEnvVars
    // (overrides EnvironmentContributingAction#buildEnvVars)
    // is applicable only for AbstractBuild.
    for(ParameterValue pv: pa.getParameters()) {
        pv.buildEnvironment(build, env);
    }
}
```

Anyway, if we migrate the *BuildSelector* to a new repo, the new signature should not have EnvVars as parameter, so this may not be needed anymore.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
alexsomai: Indeed, the parameters are something that should be bound to the selector.
The current *BuildSelector* API may not be very good, as you said. The API will be changed if moved to a separate repository.

alexsomai:
This PR is more like a Proof of Concept to prove that this plugin may work with the *BuildSelector* from the Copy Artifact plugin. 
Based on the current discussions that we are having https://groups.google.com/forum/#!topic/jenkinsci-dev/G-ZnJu3h67g, most probably the Selector will be moved to a separate repository, and the *BuildSelector* API will be changed a bit.

IMO, I don't think that this PR is worth to be merged into the master.

alexsomai:
So I've update this plugin to use the new *run-selector-plugin*. 
As it is right now, it uses https://github.com/alexsomai/run-selector-plugin/tree/rename-packages-classes
It's working pretty well, but the *run-selector-plugin* still needs more polishing.

alexsomai:
Closing in favor of #23

PR #21, opened 2016-07-12
[JENKINS-35976] Workspace cleanup

alexsomai:
Related to issue [[JENKINS-35976]](https://issues.jenkins-ci.org/browse/JENKINS-35976)

Summary of this pull request:
 - [x] Write documentation about workspace cleanup using the ws-cleanup-plugin
 - [x] Write tests for workspace cleanup using the ws-cleanup-plugin

CC @oleg-nenashev @martinda


alexsomai:
This build will fail until https://github.com/jenkinsci/ws-cleanup-plugin/pull/27 is merged and released.

oleg-nenashev:
It depends on https://github.com/jenkinsci/ws-cleanup-plugin/pull/27, I suppose

pom.xml
```diff
@@ -64,6 +64,12 @@
             <artifactId>workflow-step-api</artifactId>
             <version>${workflow.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>ws-cleanup</artifactId>
+            <version>0.30-SNAPSHOT</version>
+            <optional>true</optional>
```
oleg-nenashev: IIRC the decision was to use the dependency for test only. So maybe it should define the "test" <code>scope</code>

alexsomai: I remember that we have discussed to make it `optional`, but maybe I've misunderstood.

oleg-nenashev: CC @martinda

martinda:
Define the dependency as a test scope.

src/test/java/org/jenkinsci/plugins/ewm/TestUtil.java
```diff
@@ -47,4 +62,28 @@ public int compare(Disk disk1, Disk disk2) {
             }
         });
     }
+
+    public static void addExternalWorkspaceNodeProperty(Node node, String diskPoolRefId, DiskNode... diskNodes) {
```
oleg-nenashev: If it's public, it deserves some Javadoc.
Or maybe the entire class could be made Restricted

alexsomai: This method is used only for testing purposes. 
Is there any need to restrict it in this case?

alexsomai: I've restricted the entire class for external use.

alexsomai:
I think that this is ready for merge?

oleg-nenashev:
👍

PR #23, opened 2016-07-17
[JENKINS-36742] Integrate the Run Selector plugin

alexsomai:
Related to issue [[JENKINS-36742]](https://issues.jenkins-ci.org/browse/JENKINS-36742)

Summary of this pull request:
 - [x] Integrate the Run Selector plugin
 - [x] Update documentation

Depends on https://github.com/jenkinsci/run-selector-plugin/pull/7

CC @oleg-nenashev @martinda


src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/help-selectedRun.html
```diff
@@ -0,0 +1,18 @@
+<div>
+    <p>
+        If you provide this parameter, the <i>exwsAllocate</i> step will allocate the same workspace used by the
```
martinda: ... will allocate the first workspace ...

oleg-nenashev:
Looks good to me so far

martinda:
Good work on this.

pom.xml
```diff
@@ -70,13 +70,13 @@
...
             <optional>true</optional>
         </dependency>
+
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>run-selector</artifactId>
-            <version>1.0-alpha-2-SNAPSHOT</version>
-            <optional>true</optional>
+            <version>1.0-beta-1</version>
+            <scope>test</scope>
```
alexsomai: To be discussed if this should be `<scope>test</scope>` or `<optional>true</optional>`

alexsomai: To be discussed if this should be <scope>test</scope> or <optional>true</optional>

oleg-nenashev:
👍 Needs some testing, but the code looks good

src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/help-selectedRun.html
```diff
@@ -0,0 +1,18 @@
+<div>
+    <p>
+        If you provide this parameter, the <i>exwsAllocate</i> step will allocate the first workspace used by the
```
martinda: Sorry for this late comment, but there could be an option to select the Nth allocated workspace instead of the first one. Not important at this stage.

alexsomai: Interesting feature.
When is allocated, each workspace has a unique generated id. So maybe, instead of specifying the Nth allocated workspace, the user can simply specify the id of the workspace to be selected from the upstream job.
Here is the issue to keep track of it: https://issues.jenkins-ci.org/browse/JENKINS-37079.

In another train of thoughts: if I'll integrate the fingerprint engine for each allocated workspace, this can be simplified. The user can simply specify the fingerprint id of the workspace, without the need to specify the upstream job.

PR #24, opened 2016-07-18
[JENKINS-35975] Allow configurable relative path

alexsomai:
Related to issue [[JENKINS-35975]](https://issues.jenkins-ci.org/browse/JENKINS-35975)

Summary of this pull request:
 - [x] Define workspace based on a template configured for each Disk Pool entry (Jenkins global config)
 - [x] Custom workspace for the `exwsAllocate` step, by using an extra parameter: `path`
 - [x] Write tests 
 - [x] Write documentation

CC @oleg-nenashev @martinda


src/main/java/org/jenkinsci/plugins/ewm/definitions/DiskPool.java
```diff
@@ -71,6 +84,13 @@ public FormValidation doCheckDiskPoolId(@QueryParameter String value) {
             return validateRequired(value);
         }
 
+        public FormValidation doCheckWorkspaceTemplate(@QueryParameter String value) {
```
oleg-nenashev: Good practice: Restrict Stapler methods. Otherwise they become a part of the public UI

src/main/java/org/jenkinsci/plugins/ewm/definitions/DiskPool.java
```diff
@@ -57,6 +60,16 @@ public String getDescription() {
         return description;
     }
 
+    @CheckForNull
+    public String getWorkspaceTemplate() {
+        return workspaceTemplate;
+    }
+
+    @DataBoundSetter
+    public void setWorkspaceTemplate(String workspaceTemplate) {
```
oleg-nenashev: Maybe makes sense to extend constructor. We're in the alpha stage, so no problems with binary compatibility

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -122,15 +137,112 @@ protected ExternalWorkspace run() throws Exception {
...
+     * equal to the given disk pool id parameter.
+     *
+     * @param diskPoolId the matching id that the disk pool should have
+     * @param diskPools  the list of disk pools
+     * @return the disk pool whose {@link DiskPool#diskPoolId} is equal to the given disk pool id parameter
+     * @throws IOException if there isn't find any disk pool matching the disk pool id, or
+     *                     if the disk pool doesn't have defined any {@link Disk} entries
+     */
+    @Nonnull
+    private static DiskPool findDiskPool(@Nonnull String diskPoolId, @Nonnull List<DiskPool> diskPools) throws IOException {
```
oleg-nenashev: Fine for the private method. For public API a nullable method would be preferable

oleg-nenashev:
Mostly looks good to me. One concern, "$" is a valid symbol in FS paths.
Commonly such variable resolvers use the <code>${VAR}</code> syntax only. The <code>${</code> could be a check criteria then

oleg-nenashev:
My proposals from the meeting:
* Use <code>${VAR}</code> syntax only in docs
* Add form validation for code>${</code> with unclosed brackets (and other such cases)
* Add a warning if a user uses standalone <code>$</code> symbol since it's unsafe (both FormValidation and runtime?)

oleg-nenashev:
👍

PR #25, opened 2016-07-24
[JENKINS-36897] Disk Pool Restriction Extension Point

alexsomai:
Related to issue [[JENKINS-36897]](https://issues.jenkins-ci.org/browse/JENKINS-36897)

Summary of this pull request:
- [X] Implement the DiskPoolRestriction Extension Point
- [X] Write two implementations for the DiskPoolRestriction: 
 - [X] NoDiskPoolRestriction
 - [X] WhitelistUserRestriction
- [ ] Add javadocs and help files
- [ ] Write tests
- [ ] Write documentation

CC @martinda @oleg-nenashev

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
@@ -42,8 +43,8 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>http://github.com/jenkinsci/${project.artifactId}-plugin</url>
-      <tag>HEAD</tag>
-  </scm>
+        <tag>HEAD</tag>
+    </scm>
 
     <repositories>
         <repository>
@@ -63,6 +64,13 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
             <version>${workflow.version}</version>
+        </dependency>
+
+        <!-- Temporary - needed for the UserSelector describable -->
+        <dependency>
+            <groupId>com.synopsys.arc.jenkinsci.plugins</groupId>
+            <artifactId>job-restrictions</artifactId>
+            <version>0.4</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -65,6 +66,13 @@
             <version>${workflow.version}</version>
         </dependency>
 
+        <!-- Temporary - needed for the UserSelector describable -->
+        <dependency>
+            <groupId>com.synopsys.arc.jenkinsci.plugins</groupId>
+            <artifactId>job-restrictions</artifactId>
```
oleg-nenashev: If you depend on it, maybe makes sense to implement some restriction strategies based on this plugin later. Now there is an issue with Pipeline compatibility, so it won't work correctly.

Copy-pasting of the code and restricting it maybe an option.

src/main/java/org/jenkinsci/plugins/ewm/restrictions/AbstractUserRestriction.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * @author Alexandru Somai
+ */
+public abstract class AbstractUserRestriction extends DiskPoolRestriction {
+
+    @Override
+    public boolean isAllowed(@Nonnull Run<?, ?> run, @Nonnull TaskListener listener) {
+        Cause.UserIdCause userIdCause = run.getCause(Cause.UserIdCause.class);
```
oleg-nenashev: There is a corner-case, which may impact compatibility with other plugins. In old Hudson instances/plugins there was another cause - UserCause. It's deprecated for a long time, so you can just ignore it

src/main/java/org/jenkinsci/plugins/ewm/DiskPoolRestriction.java
```diff
@@ -0,0 +1,21 @@
...
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+
+import javax.annotation.Nonnull;
+
+/**
+ * @author Alexandru Somai
+ */
+public abstract class DiskPoolRestriction extends AbstractDescribableImpl<DiskPoolRestriction> implements ExtensionPoint {
```
oleg-nenashev: Good practice: add static all() method

oleg-nenashev:
Looks good so far

src/main/java/org/jenkinsci/plugins/ewm/restrictions/AbstractUserRestriction.java
martinda: I do not understand why this class enforces the user to have triggered the build. What if the build is triggered by another build? Don't we have to look "deeper", i.e. for the initial cause of the upstream build?

src/main/resources/org/jenkinsci/plugins/ewm/restrictions/WhitelistUserRestriction/config.jelly
```diff
@@ -0,0 +1,6 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry title="${%Whitelisted users}">
+        <f:repeatableProperty field="usersList" add="${%Whitelist user}"/>
```
martinda: Should the user list be made to look more like the user list [here](https://wiki.jenkins-ci.org/download/attachments/10059796/owndb.png?version=1&modificationDate=1202251054000)?

alexsomai:
At our meeting we have discussed that there may be no need to provide an extension point for Disk Pool restriction. It can be easily done by integrating the Job Restriction plugin.
I'll be closing this in favor of #28.

PR #27, opened 2016-07-25
[JENKINS-36898] Disk Info Provider Extension Point

alexsomai:
Related to issue [[JENKINS-36898]](https://issues.jenkins-ci.org/browse/JENKINS-36898)

Summary of this pull request:
- [X] Implement DiskInfoProvider Extension Point 
- [X] Write two basic implementations of DiskInfoProvider
 - [X] NoDiskInfo
 - [X] UserProvidedDiskInfo
- [x] Discuss the fields that are needed/not needed
- [x] Write java docs
- [x] Write documentation

CC @oleg-nenashev @martinda


alexsomai:
After the discussions that we had, we have decided that at this point we do not need to specify the size of the disk. It will be just the writeSpeed and readSpeed provided.

alexsomai:
I've added java doc and help files in #29.
Also the documentation I will write it there, since they are related.

src/main/java/org/jenkinsci/plugins/ewm/utils/FormValidationUtil.java
```diff
@@ -72,10 +75,28 @@ public static FormValidation validateWorkspaceTemplate(@Nonnull String value) {
...
     }
+
+    /**
+     * Validates that the given String is a positive double.
+     *
+     * @param value the input String
+     * @return {@link FormValidation#ok()} if the input String is a positive value,
+     * {@link FormValidation#error(String)} otherwise
+     */
+    public static FormValidation validatePositiveDouble(String value) {
```
oleg-nenashev: ьфниу зкшмфеу?

oleg-nenashev: Maybe package-internal or restricted?

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
```diff
@@ -0,0 +1,45 @@
...
+    protected DiskInfoProvider(@CheckForNull Double writeSpeed, @CheckForNull Double readSpeed) {
+        this.writeSpeed = writeSpeed != null && writeSpeed >= 0 ? writeSpeed : null;
+        this.readSpeed = readSpeed != null && readSpeed >= 0 ? readSpeed : null;
+    }
+
+    /**
+     * @return all registered {@link DiskInfoProvider}s.
+     */
+    public static ExtensionList<DiskInfoProvider> all() {
+        return Jenkins.getActiveInstance().getExtensionList(DiskInfoProvider.class);
```
oleg-nenashev: A common approach is to use <code>ExtensionList#lookup</code>, because Jenkins instance maybe inaccessible in corner cases

alexsomai: I've made the changes on #29

martinda:
Looks good to me.

PR #29, opened 2016-07-30
[JENKINS-36899] Disk Allocation Strategy Extension Point

alexsomai:
Related to issue [[JENKINS-36899]](https://issues.jenkins-ci.org/browse/JENKINS-36899)

Summary of this pull request:
- [X] Transform `DiskAllocationStrategy` into an extension point
- [X] Provide `FastestReadSpeedStrategy` and `FastestWriteSpeedStrategy` allocation strategies
- [x] Write tests
- [X] Update Java doc and help files
- [x] Write documentation

There are some items to be discussed, I've marked them with `TODO`.
This PR is created on top of #27 since they are very related.

CC @oleg-nenashev @martinda


src/test/java/org/jenkinsci/plugins/ewm/steps/DiskAllocationStrategyTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Test
+    public void useAllocationStrategyInPipeline() throws Exception {
+        Disk disk1 = TestUtil.createDisk(new UserProvidedDiskInfo(0D, 1D));
+        Disk disk2 = TestUtil.createDisk(new UserProvidedDiskInfo(0D, 3D));
+        DiskPool diskPool = TestUtil.createDiskPool(disk1, disk2);
+        TestUtil.setUpDiskPools(j.jenkins, diskPool);
+
+        WorkflowRun run = TestUtil.createWorkflowJobAndRun(j.jenkins, format("" +
+                "exwsAllocate diskPoolId: '%s', strategy: [$class: 'FastestWriteSpeedStrategy']", diskPool.getDiskPoolId()));
```
alexsomai: If I want to simplify this to: 
`exwsAllocate diskPoolId: '%s', strategy: fastetWriteSpeed()` 
I need `workflow-cps` version 2.10 as test dependency.

With `workflow-cps` version 2.10, I get the following error: https://gist.github.com/alexsomai/024a8d9d911ceede13e51590ac6e1b15
Currently I haven't found a solution for it, so that's why I've used `$class` in test.

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+public abstract class DiskAllocationStrategy extends AbstractDescribableImpl<DiskAllocationStrategy> implements ExtensionPoint {
+
+    private long estimatedWorkspaceSize;
+
+    /**
+     * @return all registered {@link DiskAllocationStrategy}s.
+     */
+    @Nonnull
+    public static ExtensionList<DiskAllocationStrategy> all() {
+        return Jenkins.getActiveInstance().getExtensionList(DiskAllocationStrategy.class);
```
oleg-nenashev: NIT: <code>ExtensionList#lookup</code>

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+    public static DescriptorExtensionList<DiskAllocationStrategy, DiskAllocationStrategyDescriptor> allDescriptors() {
+        return Jenkins.getActiveInstance().getDescriptorList(DiskAllocationStrategy.class);
+    }
+
+    /**
+     * Allocates a disk from the given list. The list contains at least one {@link Disk} entry.
+     *
+     * @param disks the entries from which to allocate a disk. The list has at least one element
+     * @return the selected disk
+     * @throws IOException if any mandatory field is missing from the {@link Disk} entry
```
oleg-nenashev: I suppose the error is being thrown for any allocation issue, no?

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    @Restricted(NoExternalUse.class)
+    public long retrieveUsableSpace(Disk disk) throws IOException {
+        String masterMountPoint = disk.getMasterMountPoint();
+        if (masterMountPoint == null) {
+            String message = String.format("Mounting point from Master to the disk is not defined for Disk ID '%s'", disk.getDiskId());
+            throw new AbortException(message);
+        }
+
+        return new File(masterMountPoint).getUsableSpace();
```
oleg-nenashev: This method may throw SecurityException IIRC

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
```diff
@@ -1,45 +1,52 @@
...
 /**
+ * Contains additional information about a {@link org.jenkinsci.plugins.ewm.definitions.Disk} entry.
+ *
  * @author Alexandru Somai
  */
 @Restricted(NoExternalUse.class)
 public abstract class DiskInfoProvider extends AbstractDescribableImpl<DiskInfoProvider> implements ExtensionPoint {
 
-    @CheckForNull
-    private final Double writeSpeed;
```
oleg-nenashev: I'm fine with both approaches, but it would be useful to know the reason of selecting a non-Double one

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
```diff
@@ -1,45 +1,52 @@
...
+    private final double readSpeed;
+    private final double writeSpeed;
```
oleg-nenashev: I'm not sure Double is a good idea. Maybe int for KB/s? I doubt we need higher accuracy

src/main/java/org/jenkinsci/plugins/ewm/strategies/AbstractDiskSpeedStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+    @Nonnull
+    @Override
+    public Disk allocateDisk(@Nonnull List<Disk> disks) throws IOException {
+        long estimatedWorkspaceSize = getEstimatedWorkspaceSize();
+
+        Iterator<Disk> iterator = disks.iterator();
+        Disk candidate = null;
+
+        // try to find at least one Disk candidate that has usable space >= estimated workspace size
+        while (iterator.hasNext()) {
```
oleg-nenashev: The code could be refactored a bit. E.g. to a for loop with one check after the run

src/main/resources/org/jenkinsci/plugins/ewm/strategies/FastestReadSpeedStrategy/help-estimatedWorkspaceSize.html
```diff
@@ -0,0 +1,31 @@
...
+        It indicates the estimated size that the workspace may have.
+        It has to be in KB.
+        This can influence the Disk allocation strategy used.
+    </p>
+    <p>
+        Examples:
+    </p>
+    <p>
+        1. Let's suppose that you choose the strategy that selects the disk with the most usable space, and you have
+        provided a workspace size estimation of 100000 KB.
```
oleg-nenashev: Maybe pipeline code example?

oleg-nenashev: It's pretty hard to read long text

src/main/resources/org/jenkinsci/plugins/ewm/strategies/FastestWriteSpeedStrategy/help-estimatedWorkspaceSize.html
```diff
@@ -0,0 +1,31 @@
...
+        Optional parameter.
+        It indicates the estimated size that the workspace may have.
+        It has to be in KB.
+        This can influence the Disk allocation strategy used.
+    </p>
+    <p>
+        Examples:
+    </p>
+    <p>
+        1. Let's suppose that you choose the strategy that selects the disk with the most usable space, and you have
```
oleg-nenashev: same

src/main/resources/org/jenkinsci/plugins/ewm/strategies/MostUsableSpaceStrategy/help-estimatedWorkspaceSize.html
```diff
@@ -0,0 +1,31 @@
...
+        Optional parameter.
+        It indicates the estimated size that the workspace may have.
+        It has to be in KB.
+        This can influence the Disk allocation strategy used.
+    </p>
+    <p>
+        Examples:
+    </p>
+    <p>
+        1. Let's suppose that you choose the strategy that selects the disk with the most usable space, and you have
```
oleg-nenashev: You could reduce duplication by writing help in Jelly instead of HTML and storing Jelly on the base class level.
Storing on the upper class level should work for HTML as well

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
martinda: Does Jenkins care about the unit, e.g. Mbps, KB/s, etc? I think not. In theory the user could be told by his admin that disk A is fast, disk B is slower than A. So the user could get away by entering 10 for disk A, and 1 for disk B, regardless of actual disk speed. What matters is that a higher number means a faster disk.

In the case of Double vs another data type (e.g. Long), it is a matter of convenience. It is not easy to visually convert Gbps and Mbps in "kilobytes per second" units, and it is not easy to read/visually compare big numbers (e.g. 1000000 - how many zeros?). But I don't think we're imposing a unit.

Implemented as "Double", we're giving the user the chance to specify disk speed in any unit of their choice, Gbps, Mbps, KB/s, etc. And the user can specify things like disk A is 2 (Gbps) and disk B is 1.2 (Gbps). So do we want to tell the user "use any unit you want but be consistent otherwise comparisons won't work.", or do we want to impose a unit of speed? I prefer that we don't impose a unit. Ten years from now, maybe even Mbps will give numbers that are unread-ably long.

src/main/resources/org/jenkinsci/plugins/ewm/definitions/DiskPool/help-strategy.html
```diff
@@ -0,0 +1,15 @@
...
+        Select the default Disk allocation strategy to be used when allocating disks within this Disk Pool.
+    </p>
+    <p>
+        The allocation strategy may also be provided in the Pipeline code, thus will override the one defined in the
+        global config.
+    </p>
+    <p>
+        <b>Note:</b> Some Disk allocation strategies may require additional information to be provided.
+        For example, if you choose the fastest read/write speed strategy, then you must provide, for each disk,
+        its read/write speed. These values should be set in the <i>Disk Information</i> section, for each <i>Disk</i>
```
martinda: its read/write speeds (plural).

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
martinda: I see: it could be "double" rather than "Double", which is what the code has now.

src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/help-strategy.html
```diff
@@ -0,0 +1,12 @@
+<div>
+    <p>
+        Provide Disk allocation strategy as step parameter.
+        This will override the strategy defined in the Jenkins global config.
+    </p>
+    <p>
+        <b>Note:</b> Some Disk allocation strategies may require additional information to be provided.
+        For example, if you choose the fastest read/write speed strategy, then you must provide, for each disk,
+        its read/write speed. These values should be set in the <i>Disk Information</i> section, for each <i>Disk</i>
```
martinda: Same comment: its read/write speeds (plural).

src/main/java/org/jenkinsci/plugins/ewm/strategies/AbstractDiskSpeedStrategy.java
alexsomai: Yes this can be changed

src/main/java/org/jenkinsci/plugins/ewm/DiskInfoProvider.java
alexsomai: We have decided to use `int`

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
alexsomai: I've updated the java doc by mention the `SecurityException` too

martinda:
Looks good to me.

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
oleg-nenashev: But you do not actually handle it

oleg-nenashev:
👍 excepting one case with exception handling

src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
alexsomai: Handling it now, and I'm throwing an `AbortException`if it occurs

alexsomai:
This PR should be merged into `master`
Closing in favor of #33

PR #34, opened 2016-08-12
Info enhancement for the External Workspaces action

alexsomai:
Summary of this pull request:
- [X] Enhance the info on the `External Workspaces` action

I have added mounting point from Jenkins Master to Disk

Before
screen shot 2016-08-11 at 9 55 50 pm

After
screen shot 2016-08-11 at 10 22 42 pm

CC @oleg-nenashev @martinda

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl/index.jelly
@@ -6,16 +6,21 @@
             <j:forEach var="exws" items="${it.allocatedWorkspaces}">
                 <h2>${%Allocated External Workspace}</h2>
                 <p>
-                    ${%Disk Pool ID:}
+                    ${%Disk Pool ID}:
                     <j:out value="${exws.diskPoolId}"/>
                 </p>
                 <p>
-                    ${%Disk ID:}
+                    ${%Disk ID}:
                     <j:out value="${exws.diskId}"/>
                 </p>
                 <p>
-                    ${%Complete Path on Disk:}
+                    ${%Workspace path on} <j:out value="${exws.diskId}"/>:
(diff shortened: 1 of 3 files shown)


martinda:
Background info: I suggested this minor enhancement as it helps quickly find the complete path on disk for networks that are homogeneous (same mount points everywhere), which is the case for my use cases.

Looks good to me.

src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl/index.jelly
```diff
@@ -6,17 +6,22 @@
...
                     <j:out value="${exws.diskPoolId}"/>
                 </p>
                 <p>
-                    ${%Disk ID:}
+                    ${%Disk ID}:
                     <j:out value="${exws.diskId}"/>
                 </p>
                 <p>
-                    ${%Complete Path on Disk:}
+                    ${%Workspace path on} <j:out value="${exws.diskId}"/>:
```
oleg-nenashev: Why j:out? It potentially allows HTML injection.

src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl/index.jelly
```diff
@@ -6,17 +6,22 @@
             <j:forEach var="exws" items="${it.allocatedWorkspaces}">
                 <h2>${%Allocated External Workspace}</h2>
                 <p>
-                    ${%Disk Pool ID:}
+                    ${%Disk Pool ID}:
                     <j:out value="${exws.diskPoolId}"/>
```
oleg-nenashev: same

oleg-nenashev:
j:out usage needs to be cleaned up, but I'm fine with other code

oleg-nenashev:
👍

PR #35, opened 2016-08-12
[JENKINS-37380] Add multiple Disk Pools to Node config

alexsomai:
Related to issue [[JENKINS-37380]](https://issues.jenkins-ci.org/browse/JENKINS-37380)

Currently, the user can define only one Disk Pool in the Node config.
screen shot 2016-08-12 at 5 26 24 pm

I think that it makes sense to be able to define multiple Disk Pools in the Node Config (since in the Jenkins global config there can be defined many)

After this PR:
screen shot 2016-08-12 at 5

TODO list:
- [X] Add multiple Disk Pools to Node config
- [x] Add multiple Disk Pools to `External Workspace Templates` (from Jenkins global config)
- [x] Update documentation with new screenshots

CC @oleg-nenashev @martinda


oleg-nenashev:
Makes sense. Do you want to integrate it before the release? It may be useful to avoid writing the data migration logic

alexsomai:
Yes, I'd like to do it before the release.
Most probably I should be done today with it.

src/main/java/org/jenkinsci/plugins/ewm/nodes/DiskPoolNode.java
```diff
@@ -0,0 +1,69 @@
...
+import static hudson.Util.fixEmptyAndTrim;
+import static hudson.Util.fixNull;
+import static hudson.util.FormValidation.validateRequired;
+
+/**
+ * Describable used to define disk pool properties for each {@link hudson.model.Node} configuration.
+ *
+ * @author Alexandru Somai
+ */
+public class DiskPoolNode implements Describable<DiskPoolNode> {
```
oleg-nenashev: Maybe <code>NodeDiskPool</code>. Maybe same for <code>NodeDisk</code>?

src/main/java/org/jenkinsci/plugins/ewm/nodes/ExternalWorkspaceProperty.java
```diff
@@ -4,56 +4,36 @@
...
-        return diskPoolRefId;
+    public ExternalWorkspaceProperty(List<DiskPoolNode> diskPoolNodes) {
+        this.diskPoolNodes = fixNull(diskPoolNodes);
     }
 
     @Nonnull
-    public List<DiskNode> getDiskNodes() {
-        return diskNodes;
+    public List<DiskPoolNode> getDiskPoolNodes() {
+        return diskPoolNodes;
```
oleg-nenashev: It exposes the internal implementation (I would expect FindBugs to grumble about it). Wrapping by Collectioms.unmodifyableList() may be a good approach

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -151,25 +144,37 @@ private static ExternalWorkspaceProperty findNodeProperty(String diskPoolRefId,
...
+     *
+     * @param diskPoolRefId the disk pool reference id to be searching for
+     * @param diskPoolNodes the list of disk pools
+     * @return the Disk Pool that has the matching reference id, {@code null} otherwise
+     */
+    @CheckForNull
+    private static DiskPoolNode findDiskPoolNode(@Nonnull String diskPoolRefId,
+                                                 @Nonnull List<DiskPoolNode> diskPoolNodes) {
+        DiskPoolNode selected = null;
+        for (DiskPoolNode diskPoolNode : diskPoolNodes) {
```
oleg-nenashev: Not important since we do not expect big disk pools: "Maybe it would be useful to have a transient HashMap"

src/main/java/org/jenkinsci/plugins/ewm/nodes/DiskPoolNode.java
alexsomai: Yes, `NodeDiskPool` and `NodeDisk` are better names

src/main/java/org/jenkinsci/plugins/ewm/nodes/ExternalWorkspaceProperty.java
alexsomai: From what I've seen, FindBugs doesn't mind about this.
But I do agree, `unmodifiableList` makes sense.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: I don't expect users to have a lot of disk pools, so I don't see the need to complicate things with disk pool indexing.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -151,53 +144,65 @@ private static ExternalWorkspaceProperty findNodeProperty(String diskPoolRefId,
...
+    /**
      * Finds the disk definition from the given disks list that matches the given disk id.
      *
      * @param diskId    the disk id that the node definition should have
-     * @param diskNodes the list of available disk definitions
+     * @param nodeDisks the list of available disk definitions
      * @param nodeName  the name of the current node
      * @return the disk definition that matches the given disk id
      * @throws IOException if no disk definition was found
      *                     if the disk definition has its local root path null
```
oleg-nenashev: NIT. add dot to the end of the first sentence

oleg-nenashev:
👍 with minor comments

martinda:
Great work!

alexsomai:
I'll merge this after I find out why my builds keep failing on ci.jenkins.io https://ci.jenkins.io/job/Plugins/job/external-workspace-manager-plugin/job/master/
It's because of the _celery_ agent.

PR #36, opened 2016-08-13
Refactor test dependencies to include minimum workflow-cps 2.10

alexsomai:
I have refactored the test dependencies.

I needed minimum `workflow-cps` version 2.10 to be able to test `@Symbol` for disk allocation strategies. 
`workflow-aggregator` was not working very well with `workflow-cps` 2.10, so I have removed it, and added the needed dependencies separately.

CC @oleg-nenashev @martinda

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,8 +88,26 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
-            <artifactId>workflow-aggregator</artifactId>
-            <version>2.2</version>
+            <artifactId>workflow-job</artifactId>
+            <version>2.5</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-cps</artifactId>
+            <version>2.11</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-durable-task-step</artifactId>
+            <version>2.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-basic-steps</artifactId>
+            <version>2.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/ewm/steps/DiskAllocationStrategyTest.java
```diff
@@ -42,7 +42,7 @@ public void useAllocationStrategyInPipeline() throws Exception {
         TestUtil.setUpDiskPools(j.jenkins, diskPool);
 
         WorkflowRun run = TestUtil.createWorkflowJobAndRun(j.jenkins, format("" +
-                "exwsAllocate diskPoolId: '%s', strategy: [$class: 'FastestWriteSpeedStrategy']", diskPool.getDiskPoolId()));
+                "exwsAllocate diskPoolId: '%s', strategy: fastestWriteSpeed()", diskPool.getDiskPoolId()));
```
alexsomai: This is the main target of this PR

oleg-nenashev:
Looks good 👍

src/test/java/org/jenkinsci/plugins/ewm/steps/DiskAllocationStrategyTest.java
oleg-nenashev: Maybe you want to have tests for both old and new implementation

src/test/java/org/jenkinsci/plugins/ewm/steps/DiskAllocationStrategyTest.java
```diff
@@ -35,22 +39,26 @@ public void tearDown() {
     }
 
     @Test
-    public void useAllocationStrategyInPipeline() throws Exception {
+    @Parameters({"fastestWriteSpeed()",
+                 "[$class: 'FastestWriteSpeedStrategy']"})
+    public void useAllocationStrategyInPipeline(String strategy) throws Exception {
```
alexsomai: I've added tests for both old and new implementation by using the `Parameters` annotation from `pl.pragmatists.JUnitParam`

pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.9</version>
+        <version>2.13</version>
```
alexsomai: I've updated the parent pom version to latest as we discussed.
I did the update in this PR. I don't see the need for a separate PR for it.

oleg-nenashev: Fine for me

PR #37, opened 2016-08-16
[WIP] [PoC] Workspace locator feature

alexsomai:
This is a WIP of the workspace locator feature.
However I still have some problems, and it's not working properly

CC @oleg-nenashev @martinda


src/main/java/org/jenkinsci/plugins/ewm/steps/model/ExternalWorkspace.java
```diff
@@ -52,4 +63,21 @@ public String getMasterMountPoint() {
...
     }
+
+    @Override
+    public String getDisplayName() {
+        return String.format("External Workspace on Disk ID: %s from Disk Pool ID: %s", diskId, diskPoolId);
+    }
+
+    @Restricted(NoExternalUse.class)
+    // TODO how to reference this from index.jelly file?
+    public DirectoryBrowserSupport doWs(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, InterruptedException {
```
alexsomai: I don't know how to access this method from `mode/ExternalWorkspace/index.jelly` file.
From here:
https://github.com/alexsomai/external-workspace-manager-plugin/blob/412578a52c652f73b7c33d153425d9de743826b1/src/main/resources/org/jenkinsci/plugins/ewm/steps/model/ExternalWorkspace/index.jelly#L4

alexsomai:
This is how I managed to do the workspace browsing feature, although it's not very user friendly.

Firstly the user has to go to `Pipeline steps`, and then to the `Use external workspace : Start`:

screen shot 2016-08-17 at 4

Afterwards, by clicking the `External Workspace` link, we will be be able to browse the actual files: 

screen shot 2016-08-17 at 4 21 42 pm

This is a similar behavior of the `ws()` step.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -136,6 +140,10 @@ protected ExternalWorkspace run() throws Exception {
             throw new AbortException(message);
         }
 
+        ExternalWorkspaceActionImpl externalWorkspaceAction = new ExternalWorkspaceActionImpl(exws, flowNode);
+        flowNode.addAction(externalWorkspaceAction);
+        exws.setWorkspaceUrl(flowNode.getUrl() + externalWorkspaceAction.getUrlName());
```
alexsomai: With this implementation I manage to reference the workspace url from the general `External Workspaces` page.
To be discussed.

src/main/java/org/jenkinsci/plugins/ewm/actions/ExternalWorkspaceActionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Restricted(NoExternalUse.class)
+    public FlowNode getParent() {
+        return parent;
+    }
+
+    public DirectoryBrowserSupport doDynamic() throws IOException {
+        Queue.Executable executable = parent.getExecution().getOwner().getExecutable();
+        if (executable instanceof AccessControlled) {
```
oleg-nenashev: I don't think it makes sense to show browser otherwise

src/main/java/org/jenkinsci/plugins/ewm/actions/ExternalWorkspaceActionImpl.java
```diff
@@ -0,0 +1,95 @@
...
+            ((AccessControlled) executable).checkPermission(Item.WORKSPACE);
+        }
+        FilePath ws = getWorkspace();
```
oleg-nenashev: Likely it won't work after the build completion. Why don't you use the data you already have in WorkflowRun for your plugin?

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -53,6 +61,31 @@ public String getUrlName() {
         return "exwsAllocate";
     }
 
+    @Restricted(NoExternalUse.class)
+    public DirectoryBrowserSupport doWs(StaplerRequest req, StaplerResponse rsp, @QueryParameter String id) throws IOException, ServletException, InterruptedException {
+        ExternalWorkspace exws = null;
```
oleg-nenashev: 🐛 missing permission check

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -53,6 +61,31 @@ public String getUrlName() {
         return "exwsAllocate";
     }
 
+    @Restricted(NoExternalUse.class)
+    public DirectoryBrowserSupport doWs(StaplerRequest req, StaplerResponse rsp, @QueryParameter String id) throws IOException, ServletException, InterruptedException {
```
oleg-nenashev: CheckForNull. Maybe not the ideal behavior for Stapler methods

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -53,6 +61,31 @@ public String getUrlName() {
...
+        ExternalWorkspace exws = null;
+
+        // TODO makes sense to make #allocatedWorkspaces a HashMap
+        for (ExternalWorkspace allocatedWorkspace : allocatedWorkspaces) {
```
oleg-nenashev: I would decouple this code to the temporary <code> ExternalWorkspace locateWorkspace(String id)</code>

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
oleg-nenashev: I would rather keep a single root action for it, but this approach may work as well

src/main/java/org/jenkinsci/plugins/ewm/steps/model/ExternalWorkspace.java
oleg-nenashev: You will need a doDynamic method() in the action, which will redirect requests like RUN_URL/externalWorkspace/id to this model object

alexsomai:
This PR is more like a PoC.
Please see #44 for the working solution.

alexsomai:
Closing in favor of #44

PR #38, opened 2016-08-17
Minor changes

alexsomai:
Summary of this pull request:
- [X] Move `ExternalWorkspace` class from `..ewm.steps.model` to `..ewm.model` package
- [X] Generate random String of 32 length for the External Workspace Id (useful in the future for Fingerprints)

CC @oleg-nenashev @martinda


oleg-nenashev:
LGTM 👍

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
```diff
@@ -225,8 +226,8 @@ private String computeDefaultPathOnDisk(@Nonnull String diskId, @CheckForNull St
             throw new AbortException(message);
         }
 
-        FilePath diskFilePath = new FilePath(new File(physicalPathOnDisk));
-        return new FilePath(diskFilePath, run.getParent().getFullName() + '/' + run.getNumber()).getRemote();
+        File pathOnDisk = Paths.get(physicalPathOnDisk, run.getParent().getFullName(), String.valueOf(run.getNumber())).toFile();
```
alexsomai: I did another change here.
There wasn't any bug, but Instead of concatenating the file path with the separator, I think it's safer to use the Java APIs for this.

src/main/resources/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl/index.jelly
```diff
@@ -16,7 +16,7 @@
                 </p>
                 <p>
                     ${%Complete workspace path on} ${exws.diskId}
-                    (${%from Jenkins master}): ${exws.masterMountPoint + '/' + exws.pathOnDisk}
+                    (${%from Jenkins master}): ${exws.completeWorkspacePath}
```
alexsomai: Same here, better to have a method that returns the concatenated paths

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateExecution.java
martinda: :+1:

oleg-nenashev:
looks good to me

PR #39, opened 2016-08-19
Prepare stable release

alexsomai:
Prepare for stable release version

CC @oleg-nenashev @martinda

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ It eliminates the need to copy, archive or move files.
 
 ## Current status
 
-Currently the plugin is under development. See the links to get more info about the current status.
+Version 1.0.0 has been released!
 
 ## Design document
 
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>external-workspace-manager</artifactId>
-    <version>1.0-beta-2-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
     <packaging>hpi</packaging>
 
     <properties>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>run-selector</artifactId>
-            <version>1.0-beta-2</version>
+            <version>1.0.0</version>
```
alexsomai: The current build will fail until I release the stable Run Selector

pom.xml
```diff
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>external-workspace-manager</artifactId>
-    <version>1.0-beta-2-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
```
alexsomai: I'd choose 1.0.0 for the stable release version

README.md
```diff
@@ -21,7 +21,7 @@ It eliminates the need to copy, archive or move files.
 
 ## Current status
 
-Currently the plugin is under development. See the links to get more info about the current status.
+Version 1.0.0 has been released!
```
alexsomai: Not very sure if this is the best message for the `Current status` section.
Maybe I should remove this section?

martinda: I have seen some projects make use of a version badge. I have no idea how to do that myself.

alexsomai: Adding a badge seems like a good idea.
It shouldn't be that difficult, I've found some resources on Google, but I think that firstly I have to perform the stable release for that to work
https://badge.fury.io/

PR #41, opened 2016-08-20
Estimated workspace size conversion fix

alexsomai:
A minor conversion bug that I've found.

I didn't read very well the documentation, the method `File#getUsableSpace` returns the usable space in bytes, not in kilobytes as I thought.
Therefore, I was comparing bytes with kilobytes when the `estimatedWorkspaceSize` parameter was provided.

This is difficult to cover with automated tests, because I am not using real mounting points, to get real values. I am using only mocks.

CC @oleg-nenashev @martinda

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/DiskAllocationStrategy.java
@@ -21,6 +21,9 @@ import java.util.List;
  */
 public abstract class DiskAllocationStrategy extends AbstractDescribableImpl<DiskAllocationStrategy> implements ExtensionPoint {
 
+    private static final long MEGABYTE = 1024L * 1024L;
+
+    // estimated workspace size has to be in MB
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
Uhm, it requires careful review of the occurrences and documentation.
Maybe it would be better to divide the result of getUsableSpace() in one place

alexsomai:
So you are suggesting to have a method e.g. `getUsableSpaceInMegaBytes` that will divide the `getUsableSpace()` result?. Like `getUsableSpace()` / 1024 / 1024.? 
In this case I'll compare MB with MB.

Yes, I think it makes more sense in this way.

The occurrences are not that many yet, so I wouldn't worry about that.

alexsomai:
I've made the changes.
I am converting the disk's usable space to MB and comparing it with the estimated workspace size provided by the user.
I've tested and it should work fine.

alexsomai:
I'll merge this to include the changes in the release

PR #44, opened 2016-08-29
[JENKINS-37773] Workspace browsing 

alexsomai:
Related to issue [[JENKINS-37773]](https://issues.jenkins-ci.org/browse/JENKINS-37773)

Summary of this pull request:
- [X] Implement the workspace browsing features

CC @oleg-nenashev @martinda

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
@@ -7,6 +7,7 @@ import org.jenkinsci.plugins.ewm.model.ExternalWorkspace;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -61,5 +62,33 @@ public class ExwsAllocateActionImpl implements RunAction2 {
     @Override
     public void onLoad(Run<?, ?> run) {
         this.parent = run;
+    }
+
+    @SuppressWarnings("unused")
+    public ExternalWorkspace getDynamic(String token) {
+        return getExternalWorkspaceById(token);
+    }
+
+    /**
+     * Returns the {@link ExternalWorkspace} that has the id equal to the given {@code id}.
+     * Returns {@code null} if none was found.
+     *
+     * @param id the external workspace id
+     * @return the external workspace whose id matches the given token, {@code null} otherwise
+     */
+    @CheckForNull
+    private ExternalWorkspace getExternalWorkspaceById(String id) {
+        // TODO use transient map to cache workspaces by id
+        for (ExternalWorkspace allocatedWorkspace : allocatedWorkspaces) {
+            if (allocatedWorkspace == null) {
+                // be defensive
+                continue;
+            }
+            if (allocatedWorkspace.getId().equals(id)) {
+                return allocatedWorkspace;
+            }
+        }
+
+        return null;
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
@@ -68,6 +68,7 @@ public class ExwsAllocateActionImpl implements RunAction2 {
         this.parent = run;
     }
 
+    @Restricted(NoExternalUse.class)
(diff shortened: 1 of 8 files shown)


src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -62,4 +67,28 @@ public void onAttached(Run<?, ?> run) {
     public void onLoad(Run<?, ?> run) {
         this.parent = run;
     }
+
+    @SuppressWarnings("unused")
+    public ExternalWorkspace getDynamic(String token) {
```
oleg-nenashev: Also makes sense to have Restricted here

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -62,4 +67,28 @@ public void onAttached(Run<?, ?> run) {
...
+
+    /**
+     * Returns the {@link ExternalWorkspace} that has the id equal to the given {@code id}.
+     * Returns {@code null} if none was found.
+     *
+     * @param id the external workspace id
+     * @return the external workspace whose id matches the given token, {@code null} otherwise
+     */
+    @CheckForNull
+    private ExternalWorkspace getExternalWorkspaceById(String id) {
```
oleg-nenashev: 🐛 If you want to use such dynamic modification, you will need to take concurrency into account. All usages of workspaceMap would require synchronization

src/main/java/org/jenkinsci/plugins/ewm/model/ExternalWorkspace.java
```diff
@@ -67,4 +76,21 @@ public String getPathOnDisk() {
     public String getCompleteWorkspacePath() {
         return new File(masterMountPoint, pathOnDisk).getPath();
     }
+
+    @Override
+    public String getDisplayName() {
+        return "Workspace on " + diskId + " from " + diskPoolId;
```
oleg-nenashev: Nice2have: The string could be localizable

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
alexsomai: I have added a `synchronized` block for this in 8cb8f69.
I think it's better than having the entire method as `synchronized`.

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -83,9 +84,13 @@ public ExternalWorkspace getDynamic(String token) {
     @CheckForNull
     private ExternalWorkspace getExternalWorkspaceById(String id) {
         if (workspacesMap == null || workspacesMap.size() != allocatedWorkspaces.size()) {
```
oleg-nenashev: 🐛 workspacesMap.size() is not thread-safe as well. I would just sync the entire method

oleg-nenashev:
Maybe CacheBuilder is more preferable then manual map construction, but NIT

alexsomai:
> Maybe CacheBuilder is more preferable then manual map construction, but NIT

Interesting, I haven't used it so far. IMHO it comes with a bit of overhead, which may not be really needed in my use case. But it's just my opinion.

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -23,12 +26,14 @@
     private final List<ExternalWorkspace> allocatedWorkspaces = new LinkedList<>();
```
alexsomai: @oleg-nenashev WDYT, should I replace this `List` with a `Map`?
(of course, not actually remove the field, but make it transient, and provide the `readResolve()` method)

oleg-nenashev: I'm fine with it. IIRC there was such proposal in one of previous PRs, but finally we decided it's not important due to the low number of workspaces.

But map really make sense since you're working on fingerprint integration

alexsomai: Yes, this action currently is added only per run, so the list will not have many items in it.

But, in the future, if I'll have a different mechanism for holding the reference to all allocated workspaces, then the collection will end up getting bigger. So in that case it makes sense to have a map.

oleg-nenashev:
👍 after the quick re-check

PR #45, opened 2016-08-30
[Proposal] Rename the node's localRootPath to nodeMountPoint

martinda:
Summary of this pull request:
- Renamed localRootPath in node configurations to nodeMountPoint

The purpose of this change is to align the code with how we are calling the feature. On the master, we call this feature the "master mount point" but on the node, we call it the "local root path". So I propose we rename it to "node mount point". This PR contains changes to the code as well as to the jelly and html.

CC @alexsomai @oleg-nenashev


src/main/java/org/jenkinsci/plugins/ewm/nodes/NodeDisk.java
```diff
@@ -24,12 +24,12 @@
 public class NodeDisk implements Describable<NodeDisk> {
 
     private final String diskRefId;
-    private final String localRootPath;
+    private final String nodeMountPoint;
```
alexsomai: I think that we'll need a `readReslve()` method here to maintain backwards compatibly. Probably there aren't many users who configured the plugin yet, but still.
It's not that difficult, here are some details https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility, at the _Scenario: Rename a field_

So in our case, I think we have the following:
```java
// keep this field for backward compatibility
private transient String localRootPath;
private final String nodeMountPoint;

// ...

protected Object readResolve() {
    if (localRootPath != null) {
        nodeMountPoint = localRootPath;
    }
    return this;
}
```
But I'm not very sure if this is the right way to do it.
Maybe @oleg-nenashev can give us more details.

alexsomai:
This PR really makes sense.

src/main/java/org/jenkinsci/plugins/ewm/nodes/NodeDisk.java
```diff
@@ -38,8 +38,8 @@ public String getDiskRefId() {
     }
 
     @CheckForNull
-    public String getLocalRootPath() {
-        return localRootPath;
+    public String getNodeMountPoint() {
```
oleg-nenashev: It breaks the binary compatibility

oleg-nenashev:
We can add readResolve(), but public getter must be retained in any case. So you can add a new method and then deprecate the existing one

martinda:
I am writing a node config compatibility test using the `@LocalData` annotation. I was expecting `GlobalConfiguration.all().get(NodeDisk.class)` to work but it does not. I need to look into this deeper.

alexsomai:
:+1: 
It looks good to me. IMO it can be merged.
Thank you @martinda for this PR.
WDYT @oleg-nenashev?

src/main/java/org/jenkinsci/plugins/ewm/nodes/NodeDisk.java
```diff
@@ -24,22 +24,35 @@
 public class NodeDisk implements Describable<NodeDisk> {
 
     private final String diskRefId;
-    private final String localRootPath;
+    private transient String localRootPath;
```
oleg-nenashev: Nice2Have: add <code>@Deprecated</code>

oleg-nenashev:
Looks good

alexsomai:
Merging.
Thanks again @martinda!

PR #46, opened 2016-08-30
[JENKINS-36413] Fingerprints integration

alexsomai:
Related to issue [[JENKINS-36413]](https://issues.jenkins-ci.org/browse/JENKINS-36413)

Summary of this pull request:
- Integrate fingerprints engine

The user will be able to view each workspace on which other jobs/builds was used.
Here's a screenshot of the result:
screen shot 2016-08-31 at 12 07 42 am

CC @oleg-nenashev  @martinda


src/main/java/org/jenkinsci/plugins/ewm/facets/WorkspaceBrowserFacet.java
```diff
@@ -0,0 +1,54 @@
...
+        if (workspace == null) {
+            return "#";
+        }
+
+        Fingerprint.BuildPtr original = getFingerprint().getOriginal();
+        if (original == null) {
+            return "#";
+        }
+
+        return original.getJob().getAbsoluteUrl() + original.getRun().getSearchUrl() +
```
oleg-nenashev: If original run gets deleted, you will have no access to this workspace anymore.
Maybe it makes sense to make the browser independent from the original run. E.g. you can just access workspaces by a link like JENKINS_URL/exws/browse/WORKSPACE_ID

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -97,6 +106,25 @@ public boolean start() throws Exception {
...
+    /**
+     * TODO JAVADOC
+     * @param workspaceId
+     * @throws IOException
+     */
+    private void registerFingerprint(String workspaceId) throws IOException {
+        FingerprintMap map = Jenkins.getActiveInstance().getFingerprintMap();
+        Fingerprint f = map.get(workspaceId);
+        if (f == null) {
+            return;
```
oleg-nenashev: Should it throw exception? Seems in such case the method does not do what it was expected

src/main/java/org/jenkinsci/plugins/ewm/facets/WorkspaceBrowserFacet.java
alexsomai: Yes, you are right, the original run may be deleted.

The solution that you are proposing means that I should have a `RootAction` that holds all the allocated workspaces? There is only one concern that I have about it. I will keep adding `ExternalWorkspace` objects to a list (or map) in this `RootAction`, but currently I don't know a mechanism of removing items from that collection (when the workspace is deleted). So, in time, the `RootAction` may end up having to many items in the collection.
Or is there any other technical solution?

At this point, I am proposing to drop the workspace browsing feature from the fingerprint facet, until I'll find a better technical solution.

The user will be able to browse the workspace by clicking on the run item -> from the drop-down list it selects the _External Workspaces_.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: Exception makes sense

src/main/java/org/jenkinsci/plugins/ewm/facets/WorkspaceBrowserFacet.java
oleg-nenashev: You do not actually need to hold workspaces anywhere else, because starting from this change you will have the Fingerprint database with it's automatic removal features :)

So Root action can actually go to Fingerprint by ID and retrieve Workspace information for browsing from it

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
```diff
@@ -67,29 +62,4 @@ public void onAttached(Run<?, ?> run) {
     public void onLoad(Run<?, ?> run) {
         this.parent = run;
     }
-
-    @Restricted(NoExternalUse.class)
-    @SuppressWarnings("unused")
-    public ExternalWorkspace getDynamic(String token) {
```
alexsomai: Now that the workspace browsing is resolved from the `RootAction`, this method is no longer needed.
I will not perform a separate release with #37, so there shouldn't be any binary compatibility issues.

src/main/java/org/jenkinsci/plugins/ewm/facets/DiskStatsInfoFacet.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jenkinsci.plugins.ewm.model.ExternalWorkspace;
+
+import javax.annotation.Nonnull;
+
+/**
+ * TODO JAVADOC
+ *
+ * @author Alexandru Somai
+ */
+public class DiskStatsInfoFacet extends FingerprintFacet {
```
alexsomai: @oleg-nenashev do you think that I still need both `FingerprintFacet`s? Maybe the `WorkspaceBrowserFacet` should be enough? and I can serve the `ExternalWorkspace` object from there?

src/main/java/org/jenkinsci/plugins/ewm/actions/ExwsAllocateActionImpl.java
oleg-nenashev: It's restricted in any case

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -108,14 +107,14 @@ public boolean start() throws Exception {
 
     /**
      * TODO JAVADOC
+     *
      * @param workspaceId
      * @throws IOException
      */
     private void registerFingerprint(String workspaceId) throws IOException {
```
martinda: Would `updateFingerprint` be a better name for this function (the fingerprint is already being registered in the `ExwsAllocateExecution` class)?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
```diff
@@ -108,14 +107,14 @@ public boolean start() throws Exception {
...
-        FingerprintMap map = Jenkins.getActiveInstance().getFingerprintMap();
-        Fingerprint f = map.get(workspaceId);
+        Fingerprint f = Jenkins.getActiveInstance()._getFingerprint(workspaceId);
         if (f == null) {
-            return;
+            throw new AbortException("Couldn't find any Fingerprint for: " + workspaceId);
```
martinda: Is it useful for the user to know the full workspace path if this occurs? If yes, maybe it is a good idea to catch the exception in the `start()` method and give more info to the user. If no, then ignore this comment.

src/main/java/org/jenkinsci/plugins/ewm/facets/DiskStatsInfoFacet.java
oleg-nenashev: Yes, it may make sense since you have only one type of fingerprints

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: I do agree with the `updateFingerprint` name

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java
alexsomai: I'm not very sure how I should handle this particular case.

If this exception occurs (although it may be an edge case scenario), when the user will try to browse the workspace for this particular fingerprint id (e.g. JENKINS_ROOT/exws/browse/wrongFingerPrintId/), he will get a nice error message

screen shot 2016-09-01 at 9 56 45 pm

alexsomai:
I think that I have addressed all the suggestions.
Maybe I can merge this? Along with #44

oleg-nenashev:
@alexsomai 👍 from me

martinda:
@alexsomai :+1:

PR #48, opened 2016-10-23
[JENKINS-39198] Fix workspace allocation in parallel test execution

alexsomai:
Related to issue [[JENKINS-39198]](https://issues.jenkins-ci.org/browse/JENKINS-39198)

The plugin has some problems when trying to run parallel tests using the same workspace, especially when running on the same node. This is because when the workspace directory is created, the method used for workspace allocation is `WorkspaceList#allocate(FilePath)`. This ads a variation to the workspace if it's in use, to make it unique, e.g. _workspace@2_, _workspace@3_, etc.
IMO we don't need this unique workspace allocation, and the workspace should be created/used without this.

Here are the Pipeline's I've used for testing this:
- Upstream job, named _test-parallel_:
  https://gist.github.com/alexsomai/654beb50aa9037a85c82500207b9837d
- Downstream job, named _run-test-suite_:
  https://gist.github.com/alexsomai/81c4ba27a828eb2e5ed73b0dd1805883

CC @oleg-nenashev @martinda


martinda:
IMO this fix matches the intent of the plugin. Thanks Alex!

alexsomai:
@oleg-nenashev I know that you may be very busy, but I can have your opinion on this too, please?

oleg-nenashev:
Looks good to me 👍

PR #49, opened 2017-07-25
Add delete disk pool button to templates and nodes

martinda:
Add button to delete disk pool in template and in node configurations

Summary of this pull request:
- Add .gitignore entry for vim
- Add delete disk pool button to template in global config, and to disk pool in node configuration

CC @alexsomai @oleg-nenashev

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/ewm/nodes/NodeDiskPool/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/ewm/nodes/NodeDiskPool/config.jelly
@@ -6,4 +6,9 @@
     <f:entry>
         <f:repeatableProperty field="nodeDisks" add="${%Add Disk}" header="${%Disk}"/>
     </f:entry>
+    <f:entry>
+        <div align="right">
+            <f:repeatableDeleteButton value="${%Delete Disk Pool}"/>
+        </div>
+    </f:entry>
 </j:jelly>
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ work
 
 # IntelliJ
 .idea
+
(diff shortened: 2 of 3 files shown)


src/main/resources/org/jenkinsci/plugins/ewm/nodes/NodeDiskPool/config.jelly
```diff
@@ -6,4 +6,9 @@
     <f:entry>
         <f:repeatableProperty field="nodeDisks" add="${%Add Disk}" header="${%Disk}"/>
     </f:entry>
+    <f:entry>
+        <div align="right">
+            <f:repeatableDeleteButton value="${%Delete Disk Pool}"/>
```
oleg-nenashev: Ideally it should be in the calling jelly. In such case you would be able to use NodeDiskPool definitions in other cases if required. Example: https://github.com/jenkinsci/github-branch-source-plugin/blob/900ea989a624223c88bddb3f1fbb3a3bd2193b5f/src/main/resources/org/jenkinsci/plugins/github_branch_source/GitHubConfiguration/config.jelly#L5-L11

alexsomai:
It looks good to me :+1:

src/main/resources/org/jenkinsci/plugins/ewm/nodes/NodeDiskPool/config.jelly
martinda: If I understand correctly, the calling jelly for NodeDiskPool/config.jelly is Template/config.jelly. When I move the delete button to the Template/config.jelly locally, the Disk Pool in the Node Configuration is missing the delete button again.

PR #51, opened 2018-09-20
[DO NOT MERGE YET] A simplistic solution for configuration as code compliance

martinda:
This is a simplistic attempt at making this plugin compliant with [Configuration as Code](https://github.com/jenkinsci/configuration-as-code-plugin)

Summary of this pull request:
- Add a setter to the list of diskPools

Next: it would be a better idea to use the [PersistedList](https://javadoc.jenkins.io/hudson/util/PersistedList.html) class, I will try to provide that too.

CC @oleg-nenashev @alexso

### DIFF
--- a/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
+++ b/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
@@ -1,0 +1,9 @@
+unclassified:
+  exwsallocatestep:
+    diskPools:
+    - diskPoolId: "diskpool1"
+      disks:
+      - diskId: "disk1"
+        displayName: "disk one display name"
+        masterMountPoint: "/tmp"
+      displayName: "diskpool1 display name"
--- a/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
+++ b/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
@@ -1,9 +1,9 @@
 unclassified:
-  exwsallocatestep:
-    diskPools:
-    - diskPoolId: "diskpool1"
-      disks:
-      - diskId: "disk1"
-        displayName: "disk one display name"
-        masterMountPoint: "/tmp"
-      displayName: "diskpool1 display name"
+    exwsallocatestep:
+      diskPools:
+          - diskPoolId: "diskpool1"
+            disks:
+                - diskId: "disk1"
+                  displayName: "disk one display name"
+                  masterMountPoint: "/tmp"
+            displayName: "diskpool1 display name"
--- a/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
+++ b/src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
@@ -1,9 +1,9 @@
 unclassified:
     exwsallocatestep:
-      diskPools:
-          - diskPoolId: "diskpool1"
-            disks:
-                - diskId: "disk1"
-                  displayName: "disk one display name"
-                  masterMountPoint: "/tmp"
-            displayName: "diskpool1 display name"
+        diskPools:
+            - diskPoolId: "diskpool1"
+              disks:
+                  - diskId: "disk1"
+                    displayName: "disk one display name"
+                    masterMountPoint: "/tmp"
+              displayName: "diskpool1 display name"
--- a/pom.xml
+++ b/pom.xml
@@ -130,5 +130,11 @@
             <version>1.0.5</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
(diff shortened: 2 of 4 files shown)


martinda:
Adding @ndeloof and @ewelinawilkosz

martinda:
Note: this PR still needs a proper casc test as described in the [CasC How to test](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/PLUGINS.md#how-to-test-) guide.

martinda:
Just want to provide an update here. The failures are because the Jenkinsfile was changed to use the default plugin build method, which makes tests run on Windows, and that is where tests are failing since this plugin works with file system paths and tests were never run on Windows. I am going through the failures one at a time.

src/test/resources/org/jenkinsci/plugins/ewm/configuration-as-code.yaml
```diff
@@ -0,0 +1,9 @@
+unclassified:
+  exwsallocatestep:
+    diskPools:
+    - diskPoolId: "diskpool1"
```
ndeloof: value after `diskPools:` should have one extra indentation
```yaml
    diskPools:
      - diskPoolId: "diskpool1"
        disks:
        - diskId: "disk1"
          displayName: "disk one display name"
          masterMountPoint: "/tmp"
        displayName: "diskpool1 display name"
```

asquel:
Hi guys, any chance you merge this one ?
@ndeloof & @martinda

martinda:
@asquel Sorry, I have been busy with [GSoC](https://jenkins.io/projects/gsoc/) activities for a while. I do not know when I can finished this. I do not remember the state of things in this PR and in the other one (which was the preferred implementation until @ndeloof explained that it would not work). I'll try to find time to have a look but I cannot commit at the moment.

imaffe:
It seems to me it just need to add an extra indentation ? I will try to get this done when I finished reading Configure as Code !

martinda:
@imaffe It's more than that. The casc feature also needs a proper test. Right now the test is almost empty, and it fails.

imaffe:
> @imaffe It's more than that. The casc feature also needs a proper test. Right now the test is almost empty, and it fails.

Yeah I found that. And I am still reading the guides of how to properly test. It might take a while since this weekend I 'll do some homework ....... I created a new pull request to see if the new PR can be detected by the autotest project. I will keep adding new tests commits to that these days.

imaffe:
@martinda  I got two questions and asking for help !
1 . In src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java 
```
        //  **Why there isn't a @DataBoundSetter annotation ?** 
        public void setDiskPools(List<DiskPool> diskPools) {
            this.diskPools.clear();
            this.diskPools.addAll(diskPools);
        }
```
according to [Descriptor Global Configuration](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#step-1)
2. In src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep/java
There is also a ```private List<Template> templates = Collections.emptyList()``` variable that doesn't have a setter function, doesn't it matter ?

imaffe:
1. In the pipeline console output there is a ```Sending interrupt signal to process``` at line 72, when entering TEST phase, not sure if we wrote wrong code or other stuff. 
- TODO : is there any other JUnit log file ? 
- TODO : how to do cherry-pick git ? and how to do make a commit on top of this ? 
ps: this is a reminder for @imaffe

martinda:
@imaffe,

For
> why there isn't a DataBoundSetter annotation on the `setDiskPools` method in src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java?

I do not know. Try adding one and see what happens. Maybe @alexsomai remembers. Or @oleg-nenashev .

For
> There is also a `private List<Template> templates = Collections.emptyList()` variable that doesn't have a setter function in src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep/java

Good question. I am not sure. This PR is not finished, it is a work in progress. You could add it and see what happens. Sorry I cannot be of more help...

For
> In the pipeline console output there is a Sending interrupt signal to process at line 72, when entering TEST phase, not sure if we wrote wrong code or other stuff.

I ran the tests locally, I am not seeing that. I would need more info to reproduce. If you still have the problem you should post the log to a gist so I can look at the full log. Also include the commands you type.

For
> how to cherry-pick (and other git questions)

Google is your friend, but try to stick to command line operations, it's the best way to learn. Cherry-pick copies a commit (or a range of commits) to the current HEAD (current branch). After cherry picking, you simply continue to add and commit. They will "go on top". Use gitk to visualize. I have a git alias set to `log --graph --abbrev-commit --oneline --decorate --all`, I find it useful for visualization too. To configure an alias: `git config alias.lga 'log --graph --abbrev-commit --oneline --decorate --all'`. Then type `git lga` to call the alias.

martinda:
This is replaced by #68 . Closing.

PR #52, opened 2018-09-20
casc compatibility

ndeloof:


src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -101,14 +102,9 @@ public DescriptorImpl() {
...
         @Override
         public boolean configure(StaplerRequest req, JSONObject formData) throws FormException {
-            diskPools = req.bindJSONToList(DiskPool.class, formData.get("diskPools"));
+            req.bindJSON(this, formData);
             save();
-            return super.configure(req, formData);
-        }
-
-        @Nonnull
-        public List<DiskPool> getDiskPools() {
```
ndeloof: you might want to keep this for backward compatibility

ndeloof: based on CI feedback, you have to :)

martinda:
The pom file for this PR is out of date for CasC. I am closing this PR in favor of #55 which is build on top of an up to date pom file.

PR #53, opened 2018-09-20
[DO NOT MERGE] casc compatibility

martinda:
Configuration as code compatibility changes.

This builds on PR #52 

However, it has problems: mvn test hang.

CC @ndeloof @ewelinawilkosz


ndeloof:
you might also add a testcase to double check casc cupport, see documentation here: https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/PLUGINS.md#how-to-test-

martinda:
This PR is based on a pom file that is not up to date with the requirements of CasC, closing it in favor of #55.

PR #54, opened 2018-09-30
Feature/upgrade pom and jenkinsfile

martinda:
This PR is to upgrade the Jenkinsfile to the standard way of building plugins using JDK8. The POM is also upgraded in preparation for configuration as code (CasC) compatibility.

Summary of this pull request:
- Upgrade Jenkinsfile to run on both *nix and Windows with JDK 8
- Upgrade POM in preparation for [CasC](https://github.com/jenkinsci/configuration-as-code-plugin)
- Make tests compatible to JDK 8 and Windows.

There will be another PR focusing just on CasC.

CC @oleg-nenashev


martinda:
A few tests fail because on Windows, there is no "sh" step. The "sh" step is used in a few of the tests. I will see if I can provide the same test using a different method. I did not catch this earlier because I run Windows tests inside a git bash shell which has "sh".

PR #56, opened 2019-01-11
Feature/configuration as code compliance

imaffe:
Related to PR#51

Summary of this pull request:
- [ fixed indention error]
- [ try to use the autotest of a plugin]

CC


martinda:
@oleg-nenashev why is this PR not being built?

imaffe:
This is a pull request not complete at all. It can be marked as DO NOT MERGE but it seems I don't have the permission to do that. Also it is not built and I don't know the triggering strategy. TODO : discover the strategy of triggering.

oleg-nenashev:
Maybe it was not built due to the infra glitch. I do not see specific
reasons in the code

alexsomai:
Really odd why this PR is not being built. I've created a new dummy PR just to test if this feature is still working, and it does. I have a new build being triggered for #58 -> https://ci.jenkins.io/job/Plugins/job/external-workspace-manager-plugin/view/change-requests/job/PR-58/

@imaffe Could it be a cause that maybe you've started this branch on top of Martin's branch, containing some of his commits?
Or maybe there was a glitch as Oleg said, and creating a new PR may fix this.

imaffe:
Currently the tests can run and is fixable by change the .yaml file I suppose. However there is  another  error log from ```Crashed tests: org.jenkinsci.plugins.ewm.steps.ExwsAllocateStepTest```
gist link is https://gist.github.com/imaffe/cb42e96f75e11383f24a156e82889364

imaffe:
The current build can complete except for some test failures that I haven't get to yet.  It seems the cause for "error exit" is the config .yaml file didn't include "jenkins" entry. This will cause an interrupt. Not sure if it is an issue ? I'll try to read the implementation of JCasC since the configuration in exws requires cascading configuration ( that is diskPools and its components must be configured and is no documented in the developer's guide). This might take a while since JCasC is a big plugin as it seems :)
So happy I got the build success ! 
CC : @oleg-nenashev @alexsomai

oleg-nenashev:
@imaffe Symbols should not be needed for Pipeline steps, because you already have `getFunctionName()`. If you want to introduce a symbol there, I believe t should have the same value. Maybe it's better to do it in a separate PR anyway

imaffe:
> @imaffe Symbols should not be needed for Pipeline steps, because you already have `getFunctionName()`. If you want to introduce a symbol there, I believe t should have the same value. Maybe it's better to do it in a separate PR anyway

@oleg-nenashev  Thanks for the help.I'll try to understand the underlying mechanism of how the CasC work. It seems JCasC has multiple ways to get the name of the component it wants to configure. Anyway it might take me some time to understand what's really going on inside there.

imaffe:
I'm gonna leave this until I finished reading JCasC source code. Im kind of busy these days doing homeworks. I will try to get this done to before Feb 26th.

martinda:
@imaffe Thanks for all your work so far. Maybe we can enlist the help of the JCasC devs on their gitter chat. Try asking @ndeloof maybe?

pom.xml
```diff
@@ -130,5 +130,11 @@
             <version>1.0.5</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.0</version>
```
timja: use a newer version of configuration-as-code please,
```suggestion
            <version>1.4</version>
```

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -89,6 +90,8 @@ public DescriptorImpl getDescriptor() {
         return (DescriptorImpl) super.getDescriptor();
     }
 
+
+    @Symbol("exwsallocatestep")
```
timja: isn't this what it would be by default? i.e. normally its generated off the class name,
you can use symbol to get a more human readable name sometimes people camelcase it,
`exWsAllocateStep` 

But as Oleg said you shouldn't need this

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
```diff
@@ -40,11 +42,18 @@ public DescriptorImpl getDescriptor() {
         return (DescriptorImpl) super.getDescriptor();
     }
 
+    @Symbol("exwsstep")
```
timja: same here?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
```diff
@@ -111,6 +114,12 @@ public boolean configure(StaplerRequest req, JSONObject formData) throws FormExc
             return Collections.unmodifiableList(diskPools);
         }
 
+        @DataBoundSetter
+        public void setDiskPools(List<DiskPool> diskPools) {
+            this.diskPools.clear();
```
timja: can't you just do?
```suggestion
            this.diskPools = diskPools;
```

src/test/java/org/jenkinsci/plugins/ewm/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @author Martin d'Anjou
+ */
+public class ConfigAsCodeTest {
+
+    @ClassRule public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void should_support_configuration_as_code() throws Exception {
+        String config = ConfigAsCodeTest.class.getResource("configuration-as-code.yaml").toString(); //configuration-as-code.yml").toString();
```
timja:
```suggestion
        String config = ConfigAsCodeTest.class.getResource("configuration-as-code.yaml").toString(); 
```

src/test/java/org/jenkinsci/plugins/ewm/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,27 @@
...
+        System.out.println("config: " + config);
+        ConfigurationAsCode.get().configure(config);
+        assertTrue(true); // check plugin has been configured as expected );
```
timja: this should have an assertion that shows the plugin is configured as expected...

pom.xml
imaffe: Thanks for the reviewing ! But now i cannot proceed because of the test failure (and thats  also why i left the test case just a skeleton). Can you take a look at the error message since I have no idea what went wrong there in the latest commit. (Will it be the version )

brief descirption of the error message : unclassified: error configuring 'unclassified' with class io.jenkins.plugins.casc.impl.configurators.GlobalConfigurationCategoryConfigurator configurator

gist : https://gist.github.com/imaffe/cb42e96f75e11383f24a156e82889364

Thanks !

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
imaffe: Yeah I will correct all these stuff when I figure out what made the test wrong.

src/test/java/org/jenkinsci/plugins/ewm/ConfigAsCodeTest.java
imaffe: Oh yeah this code is not finished yet (far from complete ...) and I will add test code when the dummy test passes. Thank you !

pom.xml
timja: Its trying to call the `setDiskPools` method but the list is an `AbstractList` and its throwing an `UnsupportedOperationException` 

If you try what I suggested here it might work:
https://github.com/jenkinsci/external-workspace-manager-plugin/pull/56#discussion_r251303959

imaffe: Thanks very much ! I'll take a look at these stuff when the class is over . Thanks very much .

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
imaffe: Whoo that really worked ! Thanks so much (I was being stupid not scolling down to those specific exceptions ) Next step would be write some really useful test cases.

timja: no problem 👍

pom.xml
martinda: Wow, I learned something here. Thank you @timja.

I re-read the List documentation and some of its methods can throw the `UnsupportedOperationException`, including the [`clear()`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#clear--) method.

Here is my understanding: Since in this case the `List` is a `List` and not an extension of `List`, the `clear()` method is [optional](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#clear--) and does not have an implementation, it can throw the exception instead. If the `List` was `ArrayList` (as an example), calling `clear()` would work since `clear()` is [not optional](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#clear--) for `ArrayList`, and it has to implement the method.

Under [Collection Framework Overview](https://docs.oracle.com/javase/8/docs/technotes/guides/collections/overview.html), you can find a discussion about this topic.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
martinda: I htink that the right name would have been:

https://github.com/jenkinsci/external-workspace-manager-plugin/blob/08073fc5c2878633f5129b6d583b83c5263274a7/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java#L120-L123

imaffe:
@oleg-nenashev @martinda I'm planning to write my first unit test code in Java (yes, first time in java LOL) for the JCasC test. The draft idea is : 
1. Read in the yaml file and see if the yaml file enties match with the internal state of exws configuration.

And this will lead to following changes in code, since I don't know if they match the coding convention, i have some questions : 
1. I need to use a yaml parser, maybe SnakeYaml. Am I allowed to add a dependency in the pom like
 ```
<dependency>
    <groupId>org.yaml</groupId>
    <artifactId>snakeyaml</artifactId>
    <version>1.21</version>            
</dependency>
is this allowed ? 
```
2. The SnakeYaml might need me to create custom classes to store the data entries. So can I create some classes like DiskTest, DiskPoolTest, TemplateTest ( which looks weired  under the test source code.)

And hope you can give me some advice on some new test points or useful techniques in this. Thank you so much !

**Comment by martinda** [comment]
@imaffe When you add a dependency which is only used for testing, you need to specify the est scope `<scope>test</scope>`. See [this](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#configuration-test) example.

As for adding a test dependency on `snakeyaml`, I would like to know if that is common practice among plugins that have made this transition. It would be good to ask on the [JCasC](https://gitter.im/jenkinsci/configuration-as-code-plugin) gitter chat. It would also be good to clarify the section on [how to test](https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#configuration-test) with that information, for future developers.

**Comment by timja** [comment]
I had a look around and snakeyaml was the best option I saw.
There's also `jackson` and pojo serialisation

There's a TODO here for providing a library to make testing exported yaml easy, currently not done though:
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

**Comment by imaffe** [comment]
@timja  Is there anyone already on this now ? If not i might offer some help like writing some documentation and give some examples. Also it would be great if you can give me a link to the related issue if there is one. Thanks so much !

**Comment by timja** [comment]
I've created a github issue:
https://github.com/jenkinsci/configuration-as-code-plugin/issues/718

File: src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java [hunk 7]
```diff
@@ -0,0 +1,92 @@
...
+
+    @ClassRule public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void should_support_configuration_as_code() throws Exception {
+        String config = ConfigAsCodeTest.class.getResource("configuration-as-code.yaml").toString();
+        ConfigurationAsCode.get().configure(config);
+
+        // parse the yaml file and check if they match
+        File file  = new File(ConfigAsCodeTest.class.getResource("configuration-as-code.yaml").getFile());
```
timja: no need to read the file in twice, you read it in on line 33

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,92 @@
...
+        String configContent = new String(Files.readAllBytes(file.toPath()));
+        Yaml yaml = new Yaml();
+        Map<String, Object> map = (Map<String, Object>) yaml.load(configContent);
+
+
+        // get the jenkins rule object.
+        Jenkins jenkins = r.getInstance();
+        ExwsAllocateStep.DescriptorImpl descriptor = (ExwsAllocateStep.DescriptorImpl) jenkins.getDescriptor(ExwsAllocateStep.class);
```
timja:
```suggestion
        ExwsAllocateStep.DescriptorImpl descriptor = ExtensionList.lookupSingleton(ExwsAllocateStep.class);
```

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,92 @@
...
+        List<DiskPool> diskPoolList = descriptor.getDiskPools();
```
timja:
```suggestion
        List<DiskPool> diskPools = descriptor.getDiskPools();
```

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: I changed it using ExtensionList.lookupSingleton but when i test it says it cannot find symbol "lookupSingleton".  Don't know what's happening...

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: When i print the config on line 33 it shows the path to the config rather than the content of the config. So I read it twice because i need the content of the yaml file to call the yaml.load() . And it is strange to me that the configure() accepts a path like "/casc/configure.yaml" to do the configuration. Anyway Martin says i can just use hardcoded values so read the file (or even use the yaml) is not necessary.

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
timja: The jenkins.version of this plugin is too old,
needs something like 2.89 for that 😢 

It can be updated separately and more modern APIs could be used then

you should be able to use getDescriptorByType and not have to cast it I think

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
timja: ok in that case you can extract the first call to `ConfigAsCodeTest.class.getResource("configuration-as-code.yaml")` to a variable and then you would do 
`configAsCodeResource.toString()`
and
` new File(configAsCodeResource.getFile())` on it

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: This time would be OK now ! Thanks a lot !

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
```diff
@@ -2,12 +2,14 @@
 
 import hudson.Extension;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
```
timja: Looks like an unused import?

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,92 @@
...
+import hudson.ExtensionList;
+import jenkins.model.Jenkins;
+import org.jenkinsci.plugins.ewm.definitions.Disk;
+import org.jenkinsci.plugins.ewm.definitions.DiskPool;
+import org.jenkinsci.plugins.ewm.steps.ExwsAllocateStep;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.*;
```
timja: not sure what the practice is here in this repo but normally * imports are discouraged as you don't know what's actually in use https://stackoverflow.com/a/147461

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,92 @@
...
+        // assertion
+        assertThat(diskPool.getDiskPoolId(), is("diskpool1"));
+        assertThat(diskPool.getDisplayName(), is("diskpool1 display name"));
+        assertThat(diskPool.getDisks().get(0).getDiskId(), is("disk1"));
+        assertThat(diskPool.getDisks().get(0).getDisplayName(), is("disk one display name"));
+        assertThat(diskPool.getDisks().get(0).getMasterMountPoint(), is("/tmp"));
+    }
+
+    @Test
+    public void should_be_backward_compatible() throws Exception {
```
timja: It looks like this is testing the same thing as the other test to me?
If so it should only be added when a backwards compatible test is needed

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,92 @@
...
+        String config = ConfigAsCodeTest.class.getResource("obsolete-configuration-as-code.yaml").toString();
+        System.out.println("config: " + config);
```
timja: if it is needed this println should be removed

imaffe: Looks like i got a lot to learn about the work flow. Sorry for the bad coding behaviour and i'll try to improve next time. I'm going to bed now and will modify them tomorrow. Thanks Tim!

timja: no worries you've done good, 
the println is useful when developing but can make test output hard to read if they're left in all over the place

imaffe:
@timja the build failed because of some FindBugs found a bug in a file I haven't modifed after I upgraded the jenkins.version from 2.60.3 to 2.89 in the pom.xml. A question is should i use 2.89.3 instead or just 2.89 (I don't know where to find the latest jenkins version or how the version number is arranged.)

timja:
Use this when selecting a version of core:
http://stats.jenkins.io/pluginversions/external-workspace-manager.html
It tells you number of plugin installs by plugin version and jenkins version
If you hover over a square it'll give you the percentage of people who are using that jenkins version or above

i.e. 2.89.1 has 92% of people using that version or above

The sections with blue ish background are LTS versions which should be targeted by plugins

imaffe:
@martinda @oleg-nenashev @timja Something weird happens. In the lastest few builds (after I upgraded the jenkins version) the FindBugs detect a bug in a file I never modified. Don't know what made the change and caused the failure. log link can be found : https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fexternal-workspace-manager-plugin/detail/PR-56/11/pipeline

martinda:
@imaffe The problem discovered by FindBugs existed before then, but was never detected. No code is 100% perfect. I suggest you break down that line into three lines to see which call causes FindBug to report that error. Then it should be easier to deal with the NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE. Perhaps one of the calls may return null, and that is what FindBugs is highlighting. One solution might be to deal with the null with a conditional that throws an exception when null is detected.

alexsomai:
Opened this code in my IDE, and most probably it complains because `nodeDisk.getNodeMountPoint()` can be null. Actually, the null check for that reference is performed in the `findNodeDisk` method, at line 225 and FindBugs is not aware of this: https://github.com/jenkinsci/external-workspace-manager-plugin/blob/3e096674102a60a6febdf59619e44c4c2018666a/src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsExecution.java#L225

What we can do is duplicate this check right before line 92:
```java
if (nodeDisk.getNodeMountPoint() == null) {
    String message = format("The Node '%s' config does not have defined any node mount point for Disk Ref ID '%s'", node.getDisplayName(), exws.getDiskId());
}
```

Although personally I'd avoid having such duplication. Is there any other way to tell FindBugs that _"hey, I've already checked for null"_?

timja:
Findbugs is correct here, line 225 is only conditionally executed, when template isn't null
So only some code branches are protected

imaffe:
This makes sense to me and I 'll correct them in the weekends. It is interesting that FindBugs can "probably" find a bug LOL. Thanks for the advice !

alexsomai:
This FindBugs starts to annoy me, so I've checked how we could fix it. And I could say, this is the most non-sense fix I've ever made. I just had to initialize the values in different variables. No additional null checks required.

https://github.com/jenkinsci/external-workspace-manager-plugin/pull/59/files

@imaffe if you revert the last changes you've made, and just copy the lines from the mentioned PR and update them on your branch, it should fix it.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
```diff
@@ -45,6 +46,11 @@ public DescriptorImpl getDescriptor() {
 
         private List<Template> templates = Collections.emptyList();
 
+
```
alexsomai: Sorry, my OCD is talking :) so I'd prefer to have only one blank line between the statements.

It seems that here are a total of two blank lines between the statements. Can you please remove one line? 

And add a new blank line after the end of `setTemplates` method, just before `public DescriptorImpl() {`?

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,70 @@
...
+ * Test for Configuration As Code Compatibility.
+ *
+ * @author Martin d'Anjou
+ */
+public class ConfigAsCodeTest {
+
+    @ClassRule public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void should_support_configuration_as_code() throws Exception {
```
alexsomai: Not strong opinion here, but what if we rename all the test methods to be camelCase? Like `shouldSupportConfigurationAsCode()` and so on? Just to be consistent with other tests among the plugin.

imaffe:
@alexsomai Looks like it won't work. Should I add several 'if' null check in before it ? I want to see what will happen.
Also I noticed that there is a place I used ```import java.io.*``` I'll fix it in next commit.

alexsomai:
@imaffe sure, let's add some null checks see what happens. Maybe something dummy in the first place:

```java
if (nodeMountPoint == null) {
    throw new AbortException();
}
if (channel == null) {
    throw new AbortException();
}
```

We can figure out the exception messages once the bug gets fixed.

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
```diff
@@ -11,6 +11,4 @@ unclassified:
           - diskId: "disk1"
             displayName: "disk one display name"
             masterMountPoint: "/tmp"
-        displayName: "diskpool1 display name"
-
-
+        displayName: "diskpool1 display name"
```
alexsomai: Sorry to bother you with these blank lines, it's really not a big deal, but at the **end** of the file it's ok to have **one** blank line :)
https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file
Same for the next file: obsolete-configuration-as-code.yaml

alexsomai:
👍 This looks good to me. Any other opinions?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
```diff
@@ -45,6 +46,10 @@ public DescriptorImpl getDescriptor() {
 
         private List<Template> templates = Collections.emptyList();
 
+        @DataBoundSetter
```
timja: normally setters comes after constructors and are grouped together with their getter

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class ConfigAsCodeTest {
+
+    @ClassRule public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void shouldSupportConfigurationAsCode() throws Exception {
+        URL resource = ConfigAsCodeTest.class.getResource("configuration-as-code.yaml");
+        String config = resource.toString();
+        ConfigurationAsCode.get().configure(config);
+        // get the jenkins rule object.
```
timja: looks like this is 1. wrong and 2. can be removed?

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertThat(diskPool.getDisks().get(0).getMasterMountPoint(), is("/tmp"));
+    }
+
+    @Test
+    public void exportConfiguration() throws Exception {
+        Yaml yaml = new Yaml();
+
+        // get the CasC configure
+        ByteArrayOutputStream outstream = new ByteArrayOutputStream();
+        ConfigurationAsCode.get().export(outstream);
```
timja: I'm slightly confused how this is working, you aren't calling `.configure` in this test, if you print the maps out do they contain what you expect?

imaffe: I read the instructions on https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md   (scroll down to the bottom) and the example seems did not use any configure... And the exported map do match the yaml file ```unclassified`` entry. Not sure if i still need to call the .configure()

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: Sorry but I don't know which line is wrong or why. Can you explain more specifically ? Thanks ! 
(If you mean the config is a absolute path like "/path/to/configure.yaml" and not the content of a yaml file, I'm curious about this but it's how the PLUGIN instructions ask me to do so LOL.)

timja: i mean this line 

> // get the jenkins rule object

The line after isn't referring to the rule object

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
timja: on the first line it says: `/** Setup jenkins to use plugin */`

i.e. configure it appropriately first either through JCASC or java api

Weird, I wonder if it's because its a class rule and its reusing the same jenkins config as the previous config.

If you run just this test does it pass?

imaffe: Yes it passed (actually I checked the content of both files and the exported yaml file do contain the ```unclassified``` entry``` But i 'll double check it in case I misunderstood the console output.

pom.xml
```diff
@@ -130,5 +129,17 @@
...
         </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
```
oleg-nenashev: It should not be needed, because snakeyaml comes from the plugin. Why did you need it?

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
```diff
@@ -45,6 +46,10 @@ public DescriptorImpl getDescriptor() {
```
oleg-nenashev: Add `@Symbol` above?

pom.xml
martinda: I'd like to move this ahead a little bit, so I have removed this dependency and adjusted the import statement.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
martinda: I added `@Symbol("exws")`, was this the right way to do it?

martinda:
I do not know why commit 59a4c80 failed while 377351f passed.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
alexsomai: I'm a bit confused, wasn't discussed previously that we don't necessarily need `@Symbol`?

> @imaffe Symbols should not be needed for Pipeline steps, because you already have `getFunctionName()`. If you want to introduce a symbol there, I believe t should have the same value. Maybe it's better to do it in a separate PR anyway

martinda: The documentation of the `Structs` plugin where `@Symbol` comes from is not clear to me. I find nothing that says the `getFunctionName()` can be used instead of the annotation.
 * [Wiki](https://wiki.jenkins.io/display/JENKINS/Structs+plugin)
* [README](https://github.com/jenkinsci/structs-plugin)
* [Symbol.java](https://github.com/jenkinsci/structs-plugin/blob/master/annotation/src/main/java/org/jenkinsci/Symbol.java)
* [Javadoc](https://javadoc.jenkins.io/plugin/structs/) => does not even mention `@Symbol`

imaffe:
> @imaffe Symbols should not be needed for Pipeline steps, because you already have `getFunctionName()`. If you want to introduce a symbol there, I believe t should have the same value. Maybe it's better to do it in a separate PR anyway

Oleg and Tim says we won't need this. But I didn't found related documentations either. and without @Symbol it worked fine. Interesting

timja:
the symbol by default is based off the descriptor class name, with or without you've mapped it to the same value, (pretty sure, may be wrong though...)

timja:
I think you've accidentally committed some kind of swap file:
`src/main/java/org/jenkinsci/plugins/ewm/steps/.ExwsStep.java.swo`

martinda:
I removed the accidental swap file and added the pattern to my `~/.config/git/ignore` file.

alexsomai:
I'd like to have this going through. Is there anything else that should be done? Can each of the reviewers please mark the discussions as _Resolved_, if that's the case? Or maybe add additional comments if needed.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
martinda: @oleg-nenashev We are confused. At first you said we do not need `@Symbol`, then you said to add it. Can you please clarify?

oleg-nenashev: Maybe `exwsConfiguration` or so, to be decided by @martinda @imaffe @alexsomai

martinda: Thanks @oleg-nenashev for this clarification. To summarize, the `@Symbol` is used by the JCasC plugin to name the map in the YAML file. If no `@Symbol` is used, then the YAML syntax stays as it is:
```
unclassified:
  exwsAllocateStep:
...
```
If the `@Symbol("exwsGlobalConfiguration")` is used, then the YAML would look like:
```
unclassified:
  exwsGlobalConfiguration:
...
```
The name of the step in the Jenkinsfile Scripted Pipeline remains determined by the value returned by `getFunctionName()` in the descriptor. It is a bit hard to wrap my head around the fact that the step definition classes contain a descriptor that is used for the global configuration. I guess that's the Jenkins architecture.

martinda: @imaffe @alexsomai I would recommend using `@Symbol(exwsGlobalConfiguration)`. But I am also okay with `@Symbol(exwsConfiguration)`.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep.java
martinda: Let's resolve this conversation in favor of https://github.com/jenkinsci/external-workspace-manager-plugin/pull/56#discussion_r257254664

martinda: In light of https://github.com/jenkinsci/external-workspace-manager-plugin/pull/56#discussion_r257254664 I suggest we use `@Symbol(exwsAllocateConfiguration)` here.

src/main/java/org/jenkinsci/plugins/ewm/steps/ExwsStep.java
alexsomai: @martinda I agree with you, `@Symbol(exwsGlobalConfiguration)` seems more appropriate.

timja: Yes the step configuration going into global can be quite confusing,
Quite a few plugins separate that out to make it easier to manage

some examples:
sonarqube
slack
azure-keyvault

(not suggesting to do that now though)

src/test/resources/org/jenkinsci/plugins/ewm/casc/obsolete-configuration-as-code.yaml
```diff
@@ -0,0 +1,16 @@
+jenkins:
```
martinda: Is this file needed at all?

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
```diff
@@ -1,14 +1,58 @@
...
   mode: NORMAL
+  nodeProperties:
+    - exwsNodeConfigurationDiskPools:
+        nodeDiskPools:
+          - diskPoolRefId: "master-node-id"
+            nodeDisks:
+              - diskRefId: "master-node-disk"
+                nodeMountPoint: "/tmp/master-node"
+
+# TODO [Martin]: The following code crashes JCasC for reasons I do not understand
```
martinda: Anyone knows what is wrong here?

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
```diff
@@ -1,14 +1,58 @@
 jenkins:
-  systemMessage: "Jenkins configured automatically by Jenkins Configuration as Code Plugin\n\n"
+  systemMessage: "Jenkins configured automatically by Jenkins Configuration as Code Plugin"
   numExecutors: 5
   scmCheckoutRetryCount: 2
   mode: NORMAL
+  nodeProperties:
```
martinda: How do I read the `nodeProperties` map in the unit test?

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
martinda: Could it be that the expected singleton is not a singleton?

martinda: I get this error when I comment out that code:
```
[ERROR] Tests run: 2, Failures: 1, Errors: 1, Skipped: 0, Time elapsed: 5.689 s <<< FAILURE! - in org.jenkinsci.plugins.ewm.casc.ConfigAsCodeTest
[ERROR] shouldSupportConfigurationAsCode(org.jenkinsci.plugins.ewm.casc.ConfigAsCodeTest)  Time elapsed: 0.047 s  <<< ERROR!
java.lang.NullPointerException
        at io.jenkins.plugins.casc.model.Mapping.lambda$clone$0(Mapping.java:71)
        at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1691)
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
        at io.jenkins.plugins.casc.model.Mapping.clone(Mapping.java:71)
        at io.jenkins.plugins.casc.model.Mapping.clone(Mapping.java:11)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151)
        at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
        at io.jenkins.plugins.casc.model.Sequence.clone(Sequence.java:43)
        at io.jenkins.plugins.casc.model.Sequence.clone(Sequence.java:9)
        at io.jenkins.plugins.casc.model.Mapping.lambda$clone$0(Mapping.java:71)
        at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1691)
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
        at io.jenkins.plugins.casc.model.Mapping.clone(Mapping.java:71)
        at io.jenkins.plugins.casc.model.Mapping.clone(Mapping.java:11)
        at io.jenkins.plugins.casc.model.Mapping.lambda$clone$0(Mapping.java:71)
        at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1691)
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
        at io.jenkins.plugins.casc.model.Mapping.clone(Mapping.java:71)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configureWith(ConfigurationAsCode.java:641)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configureWith(ConfigurationAsCode.java:545)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configure(ConfigurationAsCode.java:520)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configure(ConfigurationAsCode.java:503)
        at org.jenkinsci.plugins.ewm.casc.ConfigAsCodeTest.shouldSupportConfigurationAsCode(ConfigAsCodeTest.java:43)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:548)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.lang.Thread.run(Thread.java:748)

[ERROR] exportConfiguration(org.jenkinsci.plugins.ewm.casc.ConfigAsCodeTest)  Time elapsed: 0.272 s  <<< FAILURE!
java.lang.AssertionError: expected:<{exwsGlobalConfigurationDiskPools={diskPools=[{diskPoolId=diskpool1, disks=[{diskId=disk1, displayName=disk one display name, masterMountPoint=/tmp}], displayName=diskpool1 display name}]}, exwsGlobalConfigurationTemplates={templates=[{label=all, nodeDiskPools=[{diskPoolRefId=dp1, nodeDisks=[{diskRefId=dp1refid1, nodeMountPoint=/tmp/template11}, {diskRefId=dp1refid2, nodeMountPoint=/tmp/template12}]}, {diskPoolRefId=dp2, nodeDisks=[{diskRefId=dp2refid1, nodeMountPoint=/tmp/template21}]}]}]}}> but was:<null>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.jenkinsci.plugins.ewm.casc.ConfigAsCodeTest.exportConfiguration(ConfigAsCodeTest.java:93)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:548)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.lang.Thread.run(Thread.java:748)
```

src/test/resources/org/jenkinsci/plugins/ewm/casc/obsolete-configuration-as-code.yaml
imaffe: Actually it can be deleted.

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
imaffe: This would take a while for me to look at these stuffs. And my local repository cannot run local tests due to some unexpected error. (And that's why i didn't follow this PR after that break down ..... LOL )

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -42,11 +44,17 @@ public void shouldSupportConfigurationAsCode() throws Exception {
         String config = resource.toString();
         ConfigurationAsCode.get().configure(config);
 
+        ExternalWorkspaceProperty.DescriptorImpl exwsNodeProperty = ExtensionList.lookupSingleton(ExternalWorkspaceProperty.DescriptorImpl.class);
+        List<NodeDiskPool> nodeDiskPools = exwsNodeProperty.getNodeDiskPools();
```
martinda: The above code does not work because the descriptor does not have a list of `NodeDiskPools`:

https://github.com/jenkinsci/external-workspace-manager-plugin/blob/08073fc5c2878633f5129b6d583b83c5263274a7/src/main/java/org/jenkinsci/plugins/ewm/nodes/ExternalWorkspaceProperty.java#L35-L44

I don't really understand how to fix it... descriptors are still a mystery to me.

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
martinda: See below for my attempt at reading `nodeProperties`.

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
alexsomai: I think in this case we need to do some refactoring by moving the `List<NodeDiskPool> nodeDiskPools` from `ExternalWorkspaceProperty` class to the `DescriptorImpl`. I'm not very sure what will break by doing this 🤔 

I'll try to look these days over the required changes to be able to do such refactoring.

src/test/resources/org/jenkinsci/plugins/ewm/casc/configuration-as-code.yaml
alexsomai: I think it should be enough if you define the node as:

```
  nodes:
    - dumb:
        mode: NORMAL
        name: "agent1"
```

However I'm still trying to figure out how to define the `nodeProperties` 😕

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
alexsomai: So you can change the `ExternalWorkspaceProperty` like following, but I'm not sure if this is backwards compatible with previous plugin versions ❗️ 

```java
public class ExternalWorkspaceProperty extends NodeProperty<Node> {

    @DataBoundConstructor
    public ExternalWorkspaceProperty(List<NodeDiskPool> nodeDiskPools) {
        this.getDescriptor().setDiskPools(fixNull(nodeDiskPools));
    }

    @Override
    public DescriptorImpl getDescriptor() {
        return (DescriptorImpl) super.getDescriptor();
    }

    @Nonnull
    public List<NodeDiskPool> getNodeDiskPools() {
        return getDescriptor().getNodeDiskPools();
    }

    @Extension
    @Symbol("exwsNodeConfigurationDiskPools")
    public static class DescriptorImpl extends NodePropertyDescriptor {

        private List<NodeDiskPool> nodeDiskPools = Collections.emptyList();

        @Nonnull
        @Override
        public String getDisplayName() {
            return Messages.nodes_ExternalWorkspaceProperty_DisplayName();
        }

        @Nonnull
        public List<NodeDiskPool> getNodeDiskPools() {
            return Collections.unmodifiableList(nodeDiskPools);
        }

        @DataBoundSetter
        public void setDiskPools(List<NodeDiskPool> nodeDiskPools) {
            this.nodeDiskPools = nodeDiskPools;
        }
    }
}
```

martinda: It works for the test, but breaks backward compatibility. The ConfigMigrationTest fails.

martinda: I am thinking the NodeDiskPool class also needs to change. Descriptors are so mysterious to me.

martinda: Looks like NodeDisk also has to change.

alexsomai: Do we really need to change `NodeDiskPool` and `NodeDisk`? I think it should work even without changing those.

martinda: Well, I was trying to get the test to pass, and it was failing instead, that's how these recent changes started. Maybe the test needs to be written differently.

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author Martin d'Anjou
+ */
+public class ConfigAsCodeTest {
+
+    @ClassRule public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void shouldSupportConfigurationAsCode() throws Exception {
+        URL resource = ConfigAsCodeTest.class.getResource("configuration-as-code.yaml");
```
timja: if you add the tests dependency you can simplify this code:
```
    <dependency>
      <groupId>io.jenkins</groupId>
      <artifactId>configuration-as-code</artifactId>
      <version>1.8</version>
      <classifier>tests</classifier>
      <scope>test</scope>
    </dependency>
```

then you annotate the test method with:
```
@ConfiguredWithCode("configuration-as-code.yaml")
```

and change the rule to be:
```
@Rule
public JenkinsConfiguredWithCodeRule j = new JenkinsConfiguredWithCodeRule();
```

All the configuration as code setup bit will be done for you automatically

martinda:
Today, without changing the order of the lists in the yaml file, the list elements show up in a different order, so the test `get(0)` are not getting the value that `get(1)` were getting. There is probably no guarantee of order in the casc or the yaml. Not sure how I will fix that.

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
martinda: I agree. First I want to debug what's wrong with the Descriptor. I prefer to work on one problem at a time. See the commit 9d342da

martinda:
Sorry, I have completely dropped the ball on this. I have [pushed a fix for the get(0) problem](https://github.com/martinda/external-workspace-manager-plugin/commit/d6eaef4d927166c16523b8fc57b81bf8d53e65e9), but the test still fails in the same way. It's as if there is only one instance of the Descriptor and the whole list gets the same value for all elements.

imaffe:
@alexsomai    Hi Alex, I got some questions, in ```ExwsAllocateStep``` and ```ExwsStep```, the configurable items were declared inside the descriptor, and in``` ExternalWorkspaceProperty``` (under ```nodes``` folder ), the ```nodeDiskPools``` were declared in the outer class (rather than the inner descriptor class), is this designed to be like this ? I think the field in outer or inner class might affect the JCasC as well.

alexsomai:
@imaffe Hi Yufei, I think this was rather a design flaw. I don't remember the reason for defining `nodeDiskPools` in the outer class. I've also noticed this when I've looked over this branch, and indeed, it may cause some difficulties in integrating JCasC. If you move the `nodeDiskPools` inside the inner descriptor, there will be backwards compatibility issues. The solution I have thought about is on this branch, but it feels a bit _hackish_:

```    
@Nonnull
public List<NodeDiskPool> getNodeDiskPools() {
    return getDescriptor().getNodeDiskPools();
}
```

I'm open to other suggestions. :)

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: I'd say, we should not refactor the discriptor. According to my obeservation, only those descriptor with a ```configure``` method will need fields in the descriptor, and all those descriptors are global configurations. Node properties is a little bit special, but I take it as a not-top-level describable, as there were no ```configure()``` method defined previously. I also guess there would be a describable serve as the top-level who appear in the web UI.

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: For me, I think we might not read ExternalWorkspaceProperty like this. There should be a class containing all nodeProperties, and we suppose that class is NodeProperties. Then we do something like ```NodeProperties.getDescriptor().getProperties(ExternalWorkspaceProperty.class) ``` .(The syntax is wrong, but should be very same, I'm working on finding that class)

src/test/java/org/jenkinsci/plugins/ewm/casc/ConfigAsCodeTest.java
imaffe: @martinda Hi Martin, our previous direction was correct. I all push a new version tomorrow  : )

martinda: Thanks @imaffe looking forward to it.

martinda:
This is replaced by #68 . Closing.

PR #62, opened 2019-05-31
Adding New EfsDisk types and make WebUI work

imaffe:
Related to issue [[JENKINS-57679]](https://issues.jenkins-ci.org/browse/JENKINS-57679)
Summary of this pull request:

[Create a New EfsDisk class and make it configurable in webUI ]
CC @hynespm @martinda @alexsomai


src/main/java/org/jenkinsci/plugins/ewm/definitions/Disk.java
```diff
@@ -24,16 +27,17 @@
  *
  * @author Alexandru Somai
  */
-public class Disk implements Describable<Disk> {
+public abstract class Disk extends AbstractDescribableImpl<Disk> {
```
alexsomai: There is only one concern that I have about this approach, which is Backwards Compatibly. I never fully understood how to make sure it's not broken, but here is some useful info https://wiki.jenkins.io/display/JENKINS/Hint+on+retaining+backward+compatibility.

How can we make sure that we are backwards compatible with this approach?

imaffe: Well, I thought about this question before, and I think this would be the safe way to do it, as all old logic used only Disk's fields. I ll read that blog as well

imaffe:
Current blockers needs to discuss on Wednesday :
1. How to use base class config.jelly to reuse jelly code ? Should ask Oleg.
2. Should user provide security groups, subnet ids, for a new EFS ?
3. How to extend the ExternalWorkspace to add extra fields for ExwsExecution use ?
4. Should I use bash plugin or use Java built in shell command helper ?
5. How to import the AWS Java SDK dependencies ?
6. Investigate the `ProfileCredentialsProvider`, is it the IAM profile ?
CC @alexsomai @martinda @hynespm

alexsomai:
I'll try to answer what I can (and know):

**no. 1** -  Not very sure what's your use case, but wouldn't `<st:include ...` (from `xmlns:st="jelly:stapler")` do the work? https://wiki.jenkins.io/display/JENKINS/Understanding+Jelly+Tags

**no. 3** - I've looked a bit into the code, and it might be a bit difficult to extend it. I wasn't having this in mind when I've designed the class. 😞 Probably just adding the extra required fields to the class itself would be the easiest choice.

But anyway, what other fields do you need? Isn't enough in the `ExwsAllocateExecution` code (or event before that), to do the pre-mount, and populate all the `ExternalWorkspace` fields?

**no. 5** - Not sure if this is what you need, but don't they have a maven repo for it? https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk

imaffe:
@alexsomai For 1 : I did used that tag, but looks it didn't call the base class's config.jelly at all, it goes straight to the derived class's config.jelly. I'd further look into this, but it is not urgent.
For 3 : I put the `creating efs` in the ExwsAllocateExecution, but for mounting, I'm not sure should we do it when allocate happens. If we decide to do all the work in ExwsAllocateExecution, life would be much easier for me as we don't need to extend ExternalWorkspace. This sounds cool !
For 5 : Well I asked someone in the ec2-plugin, and they gave me some clue already, thanks !

martinda:
@imaffe For

> 6. How to bundle plugin and install it in an actual instance

See [Distributing a plugin](https://wiki.jenkins.io/display/JENKINS/Plugin+tutorial#Plugintutorial-DistributingaPlugin). The short answer is:
```
mvn package
```
Then you can find the `.hpi` file in the target folder. Just upload it to the local Jenkins instance using the plugin manager UI (last tab you can upload a custom plugin).

imaffe:
I'm now listing all the questions before I read new plugin developing tutorials. If I find answers during my reading, I will mark them as solved. But if you find it easy to answer then I can speed up a little bit. If both of us cannot answer I will post them to the public emails.  No need to answer them very soon, I use this to track my own knowledges.

1. Is there an existing  Extension Point  for efs-plugin(our new plugin) ? If not, do I need to define one.
2. How to make a class configuration appear in the top-level global configuration (like the ``` External Workspace Definitions``` in global configuration) , I kind of remember we use extension points to track that.

CC @martinda @alexsomai @hynespm

alexsomai:
1. Not sure I understand correctly. Do we even need an `ExtensionPoint`?
2. If I recall correctly, if you define `global.jelly`, that should do the trick. See `external-workspace-manager-plugin/src/main/resources/org/jenkinsci/plugins/ewm/steps/ExwsAllocateStep/global.jelly`

alexsomai:
@imaffe do you think that we should still keep this PR?

PR #65, opened 2019-06-02
update jenkins.version to LTS, from 2.60.3 to 2.164.3

based2:
https://jenkins.io/changelog-stable/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
     <packaging>hpi</packaging>
 
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.164.3</jenkins.version>
         <java.level>8</java.level>
 
         <argLine>-Xmx1024m</argLine>


alexsomai:
I've also noticed your other PR about version updates. What do you think if you do all these updates in one single PR, instead of multiple ones?

based2:
@alexsomai  Better history, able to test for each change.

alexsomai:
There are some bugs in the CI identified by FindBugs with this upgrade, so I think they should be solved first.

`Failed to execute goal org.codehaus.mojo:findbugs-maven-plugin:3.0.5:check (findbugs) on project external-workspace-manager: failed with 2 bugs and 0 errors `

https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fexternal-workspace-manager-plugin/detail/PR-65/1/pipeline

alexsomai:
Also I agree with Oleg here, I think upgrading to such latest version may be a bit risky. Do we really need it?

martinda:
Plugins try to stay "behind" Jenkins as much as they can unless a vital new feature is needed. Otherwise, older versions of Jenkins that are still in the field cannot upgrade their plugins without upgrading the core itself. Not every installation can afford to upgrade the core due to interactions with other plugins. So unless this is done to support a new feature, I vote no to this PR.

alexsomai:
A Jenkins LTS bump to an older version was done in #68. Until new features are needed, I'd prefer to stay with that version for the moment.

PR #73, opened 2020-01-17
Upgrade to use jcasc test harness

timja:
JCasc had to create a new dependency that wasn't a `tests` classifier because maven doesn't allow you to bring dependencies through classifiers.

Upgrading here so that PCT works again,

Tracking:
https://github.com/jenkinsci/bom/pull/164

Note: We'll need a release for PCT please

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.50</version>
+        <version>3.55</version>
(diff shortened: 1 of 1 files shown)


alexsomai:
Thanks, LGTM 👍 

Q: The PCT release should be done before this one is merged, or it's not related?

timja:
> Thanks, LGTM 👍
> 
> Q: The PCT release should be done before this one is merged, or it's not related?

This plugin released so that compatibility testing with this plugin works.

timja:
Yes

PR #82, opened 2026-07-28
[SECURITY-3907] Require Item/Workspace permission to browse an external workspace

alexsomai:
Follow-up to #81 (released in 1.4.1). Targets 1.4.2.

## Problem

The 1.4.1 fix gated the global `/exws/browse/{workspaceId}/ws/` route on **`Item/Read`** of a job associated with the workspace. As raised in security review, that is the wrong permission:

- Jenkins has a dedicated **`Item/Workspace`** permission for reading a job's workspace, and core's own workspace browser (`AbstractProject#doWs`) gates on it — precisely because a workspace can contain secrets (checked-out credentials, generated config, artifacts).
- `Item/Read` and `Item/Workspace` are **independent** permissions (both only `impliedBy` the generic read); `Item/Read` does **not** grant `Item/Workspace`.

So a user who has `Item/Read` on the owning job — can see it in the UI, view its config and build history — but was **not** granted `Item/Workspace` could still read the external workspace files through this route, whereas core would deny them at `/job/…/ws/`.

## Fix

Gate on **`Item/Workspace`** instead. Each job associated with the fingerprint (the allocating job and any reusing job) is resolved with system privileges — so the lookup is not itself filtered by the current user's `Item/Read` (`getItemByFullName` hides unreadable items) — and the `Item/Workspace` permission is then evaluated against the current user. This mirrors core's `doWs()`. On denial the same `NoFingerprintMatch` (404) response as for a missing fingerprint is returned, so the existence of the workspace is not leaked.

## Tests

`WorkspaceBrowserPermissionTest` now covers:

- `overallReadOnlyUserCannotBrowseWorkspace` — `Overall/Read` only → denied.
- `itemReadWithoutWorkspacePermissionCannotBrowseWorkspace` — `Item/Read` on the owning job but no `Item/Workspace` → **denied** (the case missed by 1.4.1).
- `userWithWorkspacePermissionCanBrowseWorkspace` — `Item/Workspace` on the owning job → allowed.
- `adminCanBrowseWorkspace` — administrator → allowed.

Verified the `Item/Read`-without-`Workspace` test **fails** against the old `Item/Read` behavior (browse returned 200) and passes with this fix. Full suite: 112 tests, 0 failures (green on JDK 11; the CI JDK 21 failure is the pre-existing baseline issue).

## Changelog

Adds a `1.4.2` Security entry.


src/main/java/org/jenkinsci/plugins/ewm/actions/WorkspaceBrowserRootAction.java
```diff
@@ -65,56 +69,62 @@ public Object getBrowse(String workspaceId) throws IOException {
...
      */
-    private static boolean canReadJob(@Nonnull String jobFullName) {
-        try {
-            return Jenkins.getActiveInstance().getItemByFullName(jobFullName, Job.class) != null;
-        } catch (AccessDeniedException e) {
-            return false;
+    private static boolean hasWorkspacePermission(@Nonnull String jobFullName) {
+        Job<?, ?> job;
+        try (ACLContext ignored = ACL.as2(ACL.SYSTEM2)) {
+            job = Jenkins.getActiveInstance().getItemByFullName(jobFullName, Job.class);
```
daniel-beck: This means that Job/Read isn't needed anymore, which is wrong. Both should be required.

alexsomai: Fixed. Please check again.

alexsomai:
@daniel-beck @Kevin-CB any other feedback on this PR? I'd merge if there are no objections.

Kevin-CB (approved):
Confirmed locally, it's correctly fixed. 

Thanks, please release a new version containing this change.

alexsomai:
Thanks for checking!
Released under 1.4.2

